PR #1, opened 2022-11-24
Use typemax(Int32) for last character in range

fredrikekre:
I tested out JETLS a bit in neovim, but neovim crashes when trying to display a range with `typemax(Int)` as the last column. This patch uses `typemax(Int32)` instead, which should still be much larger than any line length and serve the same purpose as "end of line" column. Most likely this can/should be fixed in neovim, but seemed easier to change it here for now.


VarLad:
@fredrikekre Can you provide the arguments to run JetLS on NeoVim?

PR #11, opened 2025-03-09
add debounce/throttle mechanism for re-analysis

abap34:
Currently, analysis is simply skipped if the interval between requests is less than `analysis_interval` (which is currently set to 5 seconds).  
https://github.com/aviatesk/JETLS.jl/blob/a594087aca4e8928b8fe8faa4cb7e8afee7d3c42/src/JETLS.jl#L728

However, in many cases, it is  inconvenient that fixes to errors are not reflected until the next edit.

This PR changes the behavior so that if a event arrives within a short period, analysis is delayed until the configured interval elapses instead of being skipped like [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server/blob/bb1c105d30a916e92267ce411eed78cdf6ec2980/src/lsp-server.ts#L316), [python-lsp-server](https://github.com/python-lsp/python-lsp-server/blob/cc6d398befddb0c9b0f0a52affade3ca4fbfded4/pylsp/python_lsp.py#L439).

### DIFF
--- a/src/JETLS.jl
+++ b/src/JETLS.jl
@@ -53,12 +53,14 @@ end
 
 struct ExternalContext end
 
+const DEFAULT_DEBOUNCE_INTERVAL = 1.5
+const DEFAULT_THROTTLE_INTERVAL = 1.5
+
 struct ServerState{F}
     send::F
     workspaceFolders::Vector{URI}
     file_cache::Dict{URI,FileInfo}
     contexts::Dict{URI,Union{Set{AnalysisContext},ExternalContext}}
-    analysis_interval::Int
     root_path::Ref{String}
     root_env_path::Ref{String}
 end
@@ -68,7 +70,6 @@ function ServerState(send::F) where F
         URI[],
         Dict{URI,FileInfo}(),
         Dict{URI,Union{Set{AnalysisContext},ExternalContext}}(),
-        5,
         Ref{String}(),
         Ref{String}())
 end
@@ -349,7 +350,8 @@ function handle_DidChangeTextDocumentNotification(state::ServerState, msg::DidCh
             analysis_context.result.staled = true
         end
         # TODO support multiple analysis contexts, which can happen if this file is included from multiple different contexts
-        reanalyze_with_context!(state, first(contexts))
+
+        debounced_reanalyze_with_context!(state, first(contexts))
     end
     nothing
 end
@@ -705,6 +707,44 @@ function initiate_context!(state::ServerState, uri::URI)
     nothing
 end
 
+
+function debounce(f::Function; delay::Float64=DEFAULT_DEBOUNCE_INTERVAL)
+    timer = nothing
+    function debounced_func(args...; kwargs...)
+        if timer !== nothing
+            close(timer)
+        end
+        timer = Timer(delay) do timer
+            f(args...; kwargs...)
+            timer = nothing
+        end
+    end
+    return debounced_func
+end
+
+function throttle(f::Function; delay::Float64=DEFAULT_THROTTLE_INTERVAL)
+    last_call = 0.0
+    timer = nothing
+    function throttled_func(args...; kwargs...)
+        now = time()
+        if now - last_call ≥ delay
+            f(args...; kwargs...)
+            last_call = now
+        else
+            if timer !== nothing
+                close(timer)
+            end
+            timer = Timer(delay - (now - last_call)) do timer
+                f(args...; kwargs...)
+                last_call = time()
+                timer = nothing
+            end
+        end
+    end
+    return throttled_func
+end
+
+
 function reanalyze_with_context!(state::ServerState, analysis_context::AnalysisContext)
     analysis_result = analysis_context.result
     if !analysis_result.staled
@@ -725,8 +765,6 @@ function reanalyze_with_context!(state::ServerState, analysis_context::AnalysisC
         notify_diagnostics!(state, (
             uri => parse_error_to_diagnostics(parse_error) for (uri, parse_error) in parse_errors))
         return nothing
-    elseif time() - analysis_result.last_analysis < state.analysis_interval
-        return nothing # no update
     end
     entry = analysis_context.entry
     include_callback = IncludeCallback(state)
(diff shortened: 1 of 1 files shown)


src/JETLS.jl
```diff
@@ -349,7 +351,12 @@ function handle_DidChangeTextDocumentNotification(state::ServerState, msg::DidCh
             analysis_context.result.staled = true
         end
         # TODO support multiple analysis contexts, which can happen if this file is included from multiple different contexts
-        reanalyze_with_context!(state, first(contexts))
+
+        if time() - first(contexts).result.last_analysis < state.analysis_interval
+            debounced_reanalyze_with_context!(state, first(contexts))
+        else
+            reanalyze_with_context!(state, first(contexts))
+        end
```
aviatesk: How about generalizing this code like below?
It can more accurately achieve the debounce purpose of running the analysis once every 5 seconds at most, and I think we can also remove the `if` branch.
```suggestion
        elapsed = time() - first(contexts).result.last_analysis
        debounce(; delay=min(0.0, DEBOUNCE_INTERVAL-elapsed)) do
            reanalyze_with_context!(state, first(contexts))
        end
```

aviatesk (commented):
That's a great idea. Thanks for the suggestion.

src/JETLS.jl
abap34: Thank you for your review! 😺

First, "debounce" is a mechanism that an event is triggered only after a specified period has elapsed since the last request. This means the delay period is constant.

Conversely, "throttle" ensures events at most once every fixed interval.
When using throttle, the analysis might run while user is still typing, and by the time the analysis completes, the code may have already changed. On the other hand, debounce triggers analysis only after editing has settled. I think that is why many language servers prefer debounce over throttle for their analysis mechanism.

So my intention is to have a fixed-width delay, not to run the analysis at most once every 5 seconds.

abap34: Additionally, since debounce requires sharing closure variables to cancel previous timers, calling it every time doesn’t work well.

Handling this with a macro might be another good approach to keep code clean.

abap34: So,  I just realized that `if` branch doesn't need 🙃
It seems correct to simply always call `debounced_reanalyze_with_context!`.

abap34:
As your suggestion, I implemented both "debounce" and "throttle".

Additionally, I simplified these implementations and set the `DEFAULT_DEBOUNCE_INTERVAL` to 1.5 seconds. This setting should eventually be user-configurable and fetched from the server.

It appears that this feature can be implemented using `workspace/configuration`, so I will implement it in another PR.

src/JETLS.jl
```diff
@@ -705,6 +707,44 @@ function initiate_context!(state::ServerState, uri::URI)
...
+            f(args...; kwargs...)
+            last_call = now
+        else
+            if timer !== nothing
+                close(timer)
+            end
+            timer = Timer(delay - (now - last_call)) do timer
+                f(args...; kwargs...)
+                last_call = time()
+                timer = nothing
```
aviatesk:
```suggestion
            timer = Timer(delay - (now - last_call)) do _
                f(args...; kwargs...)
                last_call = time()
                timer = nothing
```
I think this `timer = nothing` is intended to overwrite the `timer` variable from the outer scope—am I correct?

src/JETLS.jl
```diff
@@ -705,6 +707,44 @@ function initiate_context!(state::ServerState, uri::URI)
...
+    last_call = 0.0
+    timer = nothing
+    function throttled_func(args...; kwargs...)
+        now = time()
+        if now - last_call ≥ delay
+            f(args...; kwargs...)
+            last_call = now
+        else
+            if timer !== nothing
+                close(timer)
```
aviatesk:
```suggestion
        if timer !== nothing
            close(timer)
        end
        if now - last_call ≥ delay
            f(args...; kwargs...)
            last_call = now
        else
```
We would like to close the previous timer in the `then` case too.

src/JETLS.jl
```diff
@@ -770,4 +808,6 @@ function reanalyze_with_context!(state::ServerState, analysis_context::AnalysisC
     nothing
 end
 
+debounced_reanalyze_with_context! = debounce(reanalyze_with_context!)
```
aviatesk: We'll probably start using debounce/throttle more often going forward, so instead of creating multiple debounced/throttled versions of each function, I'd prefer to directly wrap the call sites themselves (e.g., changing `reanalyze_with_context!(state, first(contexts))` to something like `debounce(reanalyze_with_context!, state, first(contexts))`, or even `@debounce reanalyze_with_context!(state, first(contexts))` as inspired by `@invokelatest`).

abap34: I agree that we should avoid unnecessarily increasing the "debounce version" or "throttle version" of a function by using macros!

One possible way to achieve this with call-site macros might be to create a global dict that holds the variable currently captured in the closure, using the function name or a user-specified value as the key. What do you think?

src/JETLS.jl
abap34: Yes, you are correct.
And I just realized that we don’t need to rewrite `timer` as `nothing` in the outer scope from the beginning. (It literally does nothing)

src/JETLS.jl
aviatesk: Ideally, I'd like the macros to only handle syntax transformations, so we can create throttle/debounce versions just by calling functions. For example:

```julia
let debounced = Dict{Any,Timer}()
    global function debounce(f, delay, args...; kwargs...)
        if haskey(debounced, f)
            close(debounced[f])
        end
        debounced[f] = Timer(delay) do _
            try
                f(args...; kwargs...)
            finally
                delete!(debounced, f)
            end
        end
        return nothing
    end
end
macro debounce(delay, ex)
    # ...
end
let args # callsite
    debounce(reanalyze_with_context, 1.5, args...)
    # or
    @debounce 1.5 reanalyze_with_context!(args...) # transformed to `debounce(reanalyze_with_contex!, 1.5, args...)`
end
```

I think something like this would be easier to use.
We could design throttling the same way, and use it like `@throttle 1.5 @debounce 1.5 reanalyze_with_context!(args...)`.

Since I'd like to merge this PR soon, we can save the macro version for later if needed.

abap34: This looks very good. For now, I've added debounce and throttle which can be delayed by call-site. f09823f

aviatesk (approved):
LGTM. Thanks for your contribution!

PR #13, opened 2025-04-23
Reorganisation of src/LSP/LSP.jl

iuliadmtru:
The `LSP.jl` needed some reorganising and breaking up into smaller files. I made the following choices:
- Keep _~type aliases,~_ `exports` and `method_dispatcher` in `LSP.jl`.
- Create a separate `utils` directory for `interface.jl` and `namespace.jl` which all components use.
- Create a separate `capabilities` directory for server and client capabilities. This is where all supported features will be as well.
- Have a separate file for each of the following:
    - _Base types_
    - Messages -- error codes and basic interfaces for messages and notifications
    - Documents -- position and range, `WorkspaceFolder` (this shouldn't really be here, but I couldn't find a better place for it; maybe rename the file?), all document and text document related interfaces.
    - Progress -- for tracking work done progress, partial results and trace.
    - Diagnostics -- basically everything containing `Diagnostic` in its name.
    - Initialisation -- everything related to initialising communication between server and client.
    - Shutdown -- shutdown and exit.

TODO:
- [x] Compare the reorganised layout to the LSP documentation, see how they organised things.
- [x] Clean up docs.

_Note:_
_Updates are marked with italics._


src/LSP/capabilities/server.jl
```diff
@@ -0,0 +1,100 @@
...
+    positionEncoding::Union{PositionEncodingKind.Ty, Nothing} = nothing
+
+    """
+    Defines how text documents are synced. Is either a detailed structure defining each
+    notification or for backwards compatibility the TextDocumentSyncKind number.
+    If omitted it defaults to `TextDocumentSyncKind.None`.
+    """
+    textDocumentSync::Union{Union{TextDocumentSyncOptions,
+                                  TextDocumentSyncKind.Ty},
+                            Nothing} = nothing
```
aviatesk: I'm not sure if I like this syntax. Maybe just make this one line?

src/LSP/messages.jl
```diff
@@ -0,0 +1,182 @@
+# ------------------------------------------------------------------------------------------
+# Errors.
```
aviatesk:
```suggestion
# Errors
# ======
```
For this kind of subsection-like comments, could you use this syntax?

aviatesk:
Generally I agree with this refactor, thanks.
My concern is that I think there are some LSP definitions that are hard to categorize as you commented about "Documents".
While I think the classification in this PR is generally quite sensible, I also think we might get confused about how to categorize things when we port new LSP definitions in the future.
(JYI this is why I used to keep all LSP definitions in one file)

But I agree that separating them might be easier to manage.
Maybe we shouldn't try to figure out how to categorize each definition ourselves. Instead, could we organize the directories/files based on how the original LSP specification categorizes them:
image

E.g. create `src/LSP/lifecycle-messages.jl` and define `ShutdownMessage` there (we can make it directory if each sub definitions are big).

That way, we would be able to understand how our LSP definitions are organized just by looking at the original LSP specification.

Julia and TypeScript probably have different rules about the order you define types. So, we might not be able to `include` the Julia LSP definitions in the exact order listed in the LSP spec. But I think if we allow for a few exceptions, the approach above should work.

src/LSP/capabilities/server.jl
iuliadmtru: I wanted to adhere to the 92 character limit mentioned in the [formatting guidelines](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#general-formatting-guidelines-for-julia-code-contributions). Wasn't sure how to do that in a nice way though. I can keep it in one line if you prefer. Or maybe you have another preference that splits the line still?

src/LSP/messages.jl
iuliadmtru: How about something like:
```
# Errors
# ======

...

# Messages
# ========

...

# Requests
# --------

...
```
?

iuliadmtru:
I did try to follow the LSP specification, but there were quite a few exceptions necessary and the code becomes hard to follow in my opinion. For example, the specification describes `Initialize`-related interfaces intercalated with client and server capabilities. (One specific example: [`InitializeResult`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult) is described before [`ServerCapabilities`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities), although `ServerCapabilities` should be defined before `InitializeResult`.) Since both `ClientCapabilities` and `ServerCapabilities` require some `Initialization*` interface, it makes sense to have everything related to initialisation after `ClientCapabilities` and `ServerCapabilities`.

When trying to add the document symbols feature, I got tangled in three or four such examples and I could not find a nice way to arrange all the components I needed to add. This layout is the most intuitive I could find, even though it does not follow the LSP spec very closely. It is also easily extensible. For example, in order to add a new feature (capability) you would add a new file containing all there is to it and include that file before `Server` and `ClientCapabilities` (you would also need to modify these, of course).

src/LSP/capabilities/server.jl
aviatesk: I usually prefer to keep to the 92-character limit, but I don't think we need to be strict about it, at least for this part of this project.

src/LSP/messages.jl
aviatesk: Yes.

aviatesk:
Can we avoid that problem by changing the order of `include`s? My suggestion is to align the file structure with the LSP specification,
and regarding the order of defining types in the Julia LSP definitions, I'm wondering if we can work around the Julia runtime system constraints by changing `include` order within LSP.jl.

iuliadmtru:
Ok, I'll try to do that.

Question on a separate topic: even though there are definitions for base types (following the [protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#baseTypes)), the entire codebase uses Julia types instead of these aliases. Why not use them?

iuliadmtru:
I reorganised everything. I tried to follow the specification closely this time.

There is one thing I don't like -- I had to split the [Lifecycle messages](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lifeCycleMessages) section into separate files, each of them being quite small. I did this because the [Register capability](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability) is necessary for some sections (Document syncronization and Server capabilities) which are themselves necessary for the [Initialize request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize).

Apart from this, I think this layout is easily extendable. You can add new features in the dedicated `*-features` folder, without other changes (I hope). (Except of course you would need to change `ServerCapabilities` and/or `ClientCapabilities`.) You were right :).

Let me know what you think of this organisation.

iuliadmtru:
I also added some missing docs here and there.

Regarding the docs, the base types' docs are wrapped in `@docs`. They are the only ones where this is done. Could we choose to either wrap everything in `@docs` or wrap nothing?

src/LSP/language-features/diagnostics.jl
```diff
@@ -0,0 +1,303 @@
...
+end
+
+"""
+Diagnostic registration options.
+
+# Tags
+- since – 3.17.0
+"""
+@interface DiagnosticRegistrationOptions @extends TextDocumentRegistrationOptions,
+DiagnosticOptions, StaticRegistrationOptions begin
```
aviatesk, line 56: It looks like indentation seems to be broken in some places?

aviatesk:
> I also added some missing docs here and there.
> 
> Regarding the docs, the base types' docs are wrapped in `@docs`. They are the only ones where this is done. Could we choose to either wrap everything in `@docs` or wrap nothing?

`@doc` should only be added where it's needed. Before, we were using a different type of macro, so it was needed in some places, but maybe it's not needed anymore.

aviatesk:
Thanks for carefully refactoring this. Once you address the comments above, it should be ready to merge. Thanks.

aviatesk (approved):
I made some minor adjustments. Thank you for completing the big refactor.

PR #16, opened 2025-05-07
Completion interface definitions

mlechu:
While working on local completions, I needed to take these from the LSP spec and do some text editor magic.  Hoping this saves you some time for implementing global completions @aviatesk 

Not sure how many of the comments from the LSP spec we want to keep (some of the ones they have are pretty unhelpful), but I'm keeping most of them for consistency.


aviatesk (approved):
Looking great, thanks.

> Not sure how many of the comments from the LSP spec we want to keep (some of the ones they have are pretty unhelpful), but I'm keeping most of them for consistency.

Yeah, I agree, they might not actually be very helpful, but I guess it doesn't hurt to have them. Thanks for taking care of that.

PR #17, opened 2025-05-08
Starting point for the document symbols feature

iuliadmtru:
Overview of the changes:
- The necessary pieces for the feature are in the `LSP/language-features` directory, including `TextDocumentClientCapabilities` which contains many other capabilities (only `documentSymbol` is visible, the others are commented).
- The request handler is in a new directory called `requests`.
- The client and the server announce that they support the feature.

Notes:
- I left `# TODO` comments where I thought another look is necessary and where implementation is missing (such as handling macro definitions and macro calls).
- The interfaces are exact replicas from the LSP specification. However, some of them and some properties are deprecated. Those could be omitted.
- Could we move all request handlers to the `requests` directory?
- General note: I noticed that the server crashes and restarts is there is a syntax error in the analysed code. This makes it annoying to edit code. Ideally, maybe, the syntax error would be signalled through a squiggly line and the server would just wait until it gets a request with valid syntax instead of crashing?


src/LSP/language-features/document-symbol.jl
```diff
@@ -0,0 +1,257 @@
+# TODO: Remove unnecessary kinds?
```
mlechu: I think we can keep them all for now.  If we end up wanting to trim down our LSP spec files, that can happen later, but now it would just be more work for you to figure out what we don't need.

mlechu:
I'm also OK with large files (e.g. some requests bundled together) for now—requests will probably share utilities, and I'm afraid of constant reorganization as we add things.   No strong opinions, though, and of course Shuhei should be the one to decide where things go.

pfitzseb:
> I noticed that the server crashes and restarts is there is a syntax error in the analysed code. This makes it annoying to edit code. Ideally, maybe, the syntax error would be signalled through a squiggly line and the server would just wait until it gets a request with valid syntax instead of crashing?

Yes! We should make it *really* hard to ever crash the server. It should [`textDocument/publishDiagnostics`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics) when it runs into a parser error -- Shuhei added some infra around this, but might be worth revisiting to make the analysis pipeline a bit clearer.

src/requests/document-symbol.jl
```diff
@@ -0,0 +1,113 @@
+function handle_DocumentSymbolRequest(state::ServerState, msg::DocumentSymbolRequest)
+    file_uri = URI(msg.params.textDocument.uri)
+    stream = state.file_cache[file_uri].parsed_stream
```
aviatesk: Guard against the case where `state.file_cache` doesn't have `file_uri` and its `parsed_stream` contains parse errors.

aviatesk:
> I'm also OK with large files (e.g. some requests bundled together) for now—requests will probably share utilities, and I'm afraid of constant reorganization as we add things. No strong opinions, though, and of course Shuhei should be the one to decide where things go.

I agree. For now, I don't think we need to be too strict about source code organization. I'm okay with separating reasonably independent features into their own files. It should be fine to put things like "completion" or "document symbol" in separate files, but I also don't mind if they can sit in the same file. We can do reorganization whenever it turns out to be useful in the future.

src/requests/document-symbol.jl
iuliadmtru: How should I check for a wrong `parsed_stream`? Should the `SyntaxTree` creation error if the parsing fails? Should it maybe return a diagnostic?

aviatesk: Check `parsed_stream.diagnostics`. But it would be helpful if we can show some outline even if the parsing is incomplete or contains some error.

iuliadmtru:
The implementation now uses `SyntaxTree` instead of `SyntaxNode`.

There are still some aspects that need attention:
- `struct` field bindings are duplicated when there is no explicit inner construction (I'm investigating this). In general, struct handling could be better I think.
- Some document symbol kinds are not specific enough. Function arguments have the same kind as local variables (`SymbolKind.Variable`). Macro definitions are considered of kind `SymbolKind.Function`.
- Macro calls are treated similarly to function calls, in the sense that they generate a `SymbolKind.Function` symbol for their name and appropriate symbols for their arguments. Once `JuliaLowering` handles macro expansions well, this should be changed.

Other notes/questions:
- The selection range now covers entire expressions, not just the identifier name. For example, the document symbol for a function definition has the same name as the function, but spans the entire function definition. I think this approach is more informative.
- `SymbolKind.Constant` could be used for identifiers defined as `const`, but should it?

src/requests/document-symbol.jl
iuliadmtru: Right now one of the following happens:
- Wrong syntax, lowering error: outline remains the same as is was previous to the erroring request (the request fails -- it could fail with a diagnostic instead of an error).
- Wrong syntax, lowering succeeds or correct syntax: outline is updated.

iuliadmtru: This means there's no outline when the first `DocumentSymbolRequest` contains a file with syntax errors. I'll try to fix that.

src/LSP/capabilities.jl
```diff
@@ -18,6 +18,8 @@
     """
     textDocumentSync::Union{Union{TextDocumentSyncOptions, TextDocumentSyncKind.Ty}, Nothing} = nothing
 
+    documentSymbolProvider::Union{Union{Bool, DocumentSymbolOptions}, Nothing} = nothing
```
aviatesk:
```suggestion
    """
    The server provides document symbol support.
    """
    documentSymbolProvider::Union{Union{Bool, DocumentSymbolOptions}, Nothing} = nothing
```
Just for consistency

src/LSP/language-features/text-document.jl
```diff
@@ -0,0 +1,195 @@
+"""
```
aviatesk: Following our rule for LSP organization, I think this file should live under `LSP/life-cycle-messages/`.

src/document-symbols.jl
```diff
@@ -0,0 +1,313 @@
...
+end
+
+# Request handler
+# ===============
+
+function handle_DocumentSymbolRequest(state::ServerState, msg::DocumentSymbolRequest)
+    file_uri = URI(msg.params.textDocument.uri)
+    file_info = get_fileinfo(state, file_uri)
+    isnothing(file_info) &&
+        # TODO: Send error/warning/diagnostic?
```
aviatesk:
```suggestion
```
No need to emit any diagnostics here. Diagnostics are emitted by the server upon `onTextDocumentChange` notification.

src/document-symbols.jl
```diff
@@ -0,0 +1,313 @@
...
+        return state.send(DocumentSymbolResponse(; id = msg.id, result = DocumentSymbol[]))
+    stream = file_info.parsed_stream
+    # TODO: Guard for malformed stream.
```
aviatesk:
```suggestion
```
From your comment, I think this case is now handled properly somehow?

test/test_document_symbols.jl
```diff
@@ -0,0 +1,156 @@
...
+                         kind::SymbolKind.Ty,
+                         start_pos,
+                         end_pos,
+                         children_len)
+    end_pos = (end_pos[1], end_pos[2] - 1)
+    return test_symbol(s, name, kind, start_pos, end_pos, children_len)
+end
+
+@testset "Simple expressions" begin
+    src = """
```
aviatesk:
```suggestion
    let src = """
```
I'd prefer wrapping each independent test cases within `let` block so that we don't have unexpected reuse of bindings from previous test cases.

src/document-symbols.jl
```diff
@@ -0,0 +1,313 @@
...
+
+"""
+Get the bindings from a `SyntaxTree` lowered up to scope resolution. Filter out internal
+bindings.
+"""
+function lower_and_get_bindings(st0::JL.SyntaxTree)
+    # Lower the syntax tree.
+    ctx1, st1 = JL.expand_forms_1(Module(), st0)
+    ctx2, st2 = JL.expand_forms_2(ctx1, st1)
+    ctx3, _ = JL.resolve_scopes(ctx2, st2)
```
aviatesk: Lowering in general can produce arbitrary errors since it may include execution of user code, so we really need to wrap it into `try`/`catch` and handle the lowering failure somehow. See completions.jl for how it should be done. Reuse the lowering routine if possible.

aviatesk:
I tried out the PR.
I tried a simple case like the following, but it seems the symbols aren't displayed hierarchically. Specifically, I think it's okay to show `x` as a symbol inside the `hello` function, but I think it should be displayed as a local binding of `hello`, not at the same level of `hello` (top level function). Also, since `Hello` is just the type of `x`, it doesn't need to be displayed as an independent symbol. Generally, think of this "document symbol" feature not as a feature that lists all symbols in the Julia program the user has open, but as a function that displays its "outline" in a way that we can grasp its structure easily.
Screenshot 2025-05-23 at 01 13 52

Also, in a case including varargs like the following, it seems an error occurs during lowering, but this is normal valid syntax, so let's make sure we can handle it properly:
```julia
function hello(x::Hello, xs::Hello...)
    a = hello(x)
    as = hello(xs...)
    return (a, as...)
end
```

aviatesk:
> * The selection range now covers entire expressions, not just the identifier name. For example, the document symbol for a function definition has the same name as the function, but spans the entire function definition. I think this approach is more informative.

Yeah, I agree.

> * `SymbolKind.Constant` could be used for identifiers defined as `const`, but should it?

This sounds sensible.

src/LSP/language-features/text-document.jl
iuliadmtru: Would it be better to append it to the `LSP/capabilities.jl` file? `TextDocumentClientCapabilities` are described right before `ClientCapabilities` in the LSP specification.

aviatesk: yeah that sounds fine too

iuliadmtru:
I included most changes we discussed. I still need to:
- [x] figure out why struct fields are duplicated,
- [x] mark `SymbolKind.Constant`s,
- [x] ~~use an explicit stack instead of the program stack,~~ I think this is not necessary, actually
- [x] update tests & add more,
- [ ] other things you might find during review.

iuliadmtru:
This should work now on a lot more basic cases. Some updates and issues:

- `[const]` nodes provide a `SymbolKind.Constant` symbol for the lhs instead of `SymbolKind.Variable`.
- Assignments currently only create symbols for identifier or `::` lhs nodes. For something like `const MyVector{T} = Array{1,T}` no symbol is created, mainly because I didn't know what the generated symbol should be or if there should actually be one.
- Another problem with assignments is that they ignore the rhs. I think this is a fair approach for most assignments, but not for chained assignments. `a = b = 2` should generate two symbols, one for `a` and one for `b`, but right now there's only a symbol for `a`. This is easily fixable for this specific case (chained assignments). However, if there are other cases where the rhs should not be ignored, a more general strategy would be better.
- Before, function and macro definitions were lowered and the information on arguments and local variables was obtained from the lowering context. This did not give specific enough outlines. Inner hierarchical information was lost. For example, inner functions would only appear by name (a local symbol with the function name would be shown), but not by structure (there would be no information regarding the inner function's local variables). This was fixed by recursively creating symbols for each expression in the function body.
- The current approach still uses the lowering context to get the arguments symbols. This is a problem when there are inner functions, because lowering creates `:argument` bindings for the inner functions' arguments as well. That causes the top-level function to have the inner functions' arguments as direct children, instead of only having the inner function symbols as children, with their arguments as their own children. I'm not sure what a good fix would be. It would be great if there would be a way to differentiate between the lowered function's arguments and inner arguments. @mlechu do you know if that's possible?
- Only some expressions are treated in a special way (definitions, assignments, `if`s, `for`s and `while`s, broadly speaking). All other expressions generate symbols for `:local` bindings. Surely there are other expressions that should be treated specially, but maybe this is enough for a start?

@aviatesk

aviatesk:
> * Assignments currently only create symbols for identifier or `::` lhs nodes. For something like `const MyVector{T} = Array{1,T}` no symbol is created, mainly because I didn't know what the generated symbol should be or if there should actually be one.

Can't we show `MyVector{T}` as the top-level symbol for that case?

> * Another problem with assignments is that they ignore the rhs. I think this is a fair approach for most assignments, but not for chained assignments. `a = b = 2` should generate two symbols, one for `a` and one for `b`, but right now there's only a symbol for `a`.

I'd like to get both symbols are shown as top-level symbols if the assignments happen at top-level. This means that we also want to get `a` and `b` shown for cases like
```julia
a = begin
    b = 42
    sin(b)
end
```

iuliadmtru:
> Can't we show MyVector{T} as the top-level symbol for that case?

Yes. I think it should be shown as a `Variable` symbol.

> I'd like to get both symbols are shown as top-level symbols if the assignments happen at top-level.

Alright.

iuliadmtru:
Another thing that I changed is the `struct` layout. Inner constructors now appear in the outline. For example:
```julia
struct St
    x::Int
    y::Int
    St(x::Int) = new(x, x)
    St(x::Int, y::Int) = new(x, y)
end
```
has the following outline:
image

Unfortunately, the current approach does not support inner constructors with type parameters because they are not independently lowerable. So for something like
```julia
struct S{T}
    x::T
    S{T}(x) where T = new(x)
end
```
the outline is just
```
> S
   x
   T
```
(Should the shown symbol name maybe be `S{T}` instead of just `S`?)

A hacky workaround would be to remove the constructor type parameter before lowering.

aviatesk:
> 

> Unfortunately, the current approach does not support inner constructors with type parameters because they are not independently lowerable. So for something like
> 
> ```julia
> struct S{T}
>     x::T
>     S{T}(x) where T = new(x)
> end
> ```
> 
> the outline is just
> 
> ```
> > S
>    x
>    T
> ```

In this case, can it be displayed as
```
> S{T}
  > x
  > S{T}
    > x
```
?

aviatesk:
@iuliadmtru Do you think you can find time to complete this PR?

iuliadmtru:
@aviatesk Sorry for the late reply! I have some time today and I'll try to wrap it up.

> In this case, can it be displayed as

```
> S{T}
  > x
  > S{T}
    > x
```

Sure. But how about showing the type parameter as a child of the struct as well?

```
> S{T}
  > x
  > T
  > S{T}
    > x
```

aviatesk:
> But how about showing the type parameter as a child of the struct as well?

Just my personal opinion, but since `S{T}` already shows the information that `T` is a type parameter, I don't think it's necessary.

aviatesk:
Since this PR was created, the LS implementation has evolved significantly, resulting in numerous conflicts and becoming stale, so we are closing it.
Some of the implementations in this PR will be useful references when implementing this feature in the future.

PR #19, opened 2025-05-12
Implement completions

mlechu:
Completions from local scopes using JuliaLowering

completions

## Implementation details

Solves a few problems:
1. Lowering doesn't look into certain forms like modules.  This implementation
   finds the greatest lowerable syntax tree containing the cursor and keeps
   track of the cursor within it
2. JuliaLowering throws away scope information almost as soon as it creates it
   (namely we want a binding->scope or scope->binding mapping).  Work around
   this by taking all bindings and filtering them.

A binding from lowering becomes a completion if all of the following are true:
- The binding isn't internal (compiler-generated)
- It belongs to a scope that contains the cursor (its nearest parent that is a
  scoped tree contains the cursor)
- It isn't shadowed by a binding of the same name in a tighter scope around the cursor
- It's declared before the cursor (or global)

I've added any bonus information I could get out of JuliaLowering
(local/global/const/argument/sparam classification, local type declarations, and
the `showprov` "documentation" popup), which is nice for demonstration purposes,
but it would make sense to replace these with more informative or reliable
information later when we have JuliaLowering and JET working together.

## Limitations
- Sorting by assumed importance is not yet implemented (requires tweaks to the client)
- Completions will only work if lowering works on in-progress code

## TODO
- Add tests
- Unexpected names show up in completions, which I have to go hunt down.
  - Types and operators become "global" completions with unhelpful provenance info
  - Some cases of internal symbols leaking through, e.g. `#fl_parse#0`
- Ignore variables currently being assigned to for the first time
- Instead of using the cursor's "greatest lowerable subtree", we could lower
  "the children of the least unlowerable subtree" without too much trouble,
  which would give us completions for the insides of modules.  I assume using
  JET is the more general solution, though.
- Improve provenance of errors that come from lowering (currently they come from 
  lisp and go through JET)


src/LSP/language-features/completions.jl
```diff
@@ -499,3 +501,159 @@ presented in the editor.
     """
     items::Vector{CompletionItem}
 end
+
+@interface CompletionClientCapabilities begin
```
aviatesk, line 515: Right now, we don't change how the LS behaves based on client capabilities, so maybe we don't need to add this yet. I think we'll need it later, for example, when we implement lazy completion. If a frontend doesn't support it, we might need to fall back to eager completion. Anyway, it's fine to add it. I'll just make a note of it for when we add new features later.

src/completions.jl
```diff
@@ -0,0 +1,261 @@
...
+import JuliaSyntax: @K_str, @KSet_str, JuliaSyntax as JS
+
+# get_fileinfo(s::ServerState, t::TextDocumentPositionParams) = get_fileinfo(s, t.textDocument)
+get_fileinfo(s::ServerState, t::CompletionParams) = get_fileinfo(s, t.textDocument)
+
+"""
+Like `Base.unique`, but over node ids, and with this comment promising that the
+lowest-index copy of each node is kept.
+"""
+function unique(sl::JL.SyntaxList)
```
```suggestion
function unique_syntax_list(sl::JL.SyntaxList)
```
It's a bit tricky, but if we don't use `Base.unique`, this defines a new `unique` generic function, which will shadow `Base.unique` in this module namespace. So, either explicitly extend `Base.unique` or just give it a different name. I'd recommend the later since this function is very specific to this package.

src/completions.jl
```diff
@@ -0,0 +1,261 @@
...
+
+    if !isnothing(binding.type)
+        label_detail = "::" * JL.sourcetext(binding.type)
+    end
+
+    if !isnothing(lb) && lb.is_called
+        label_kind = :Function
+    end
+
+    detail = sprint(JL.showprov, st)
```
aviatesk, line 229: In the future, we could potentially make this lazy using `resolveProvider` (just for performance). But performance doesn't seem to be a big worry even as it is, so it's fine for now.

aviatesk:
Great PR. Thanks.
Regarding completion for the global space, I'll quickly implement a simple version, and then we can think about what else is needed based on that.

> Sorting by assumed importance is not yet implemented (I think this requires
> negotiating with the client, and it looks like there's a default sorting logic
> in place anyway)

Does this mean that even if we sort the order returned on the server side, it's not reflected on the client side? If needed, it's fine to edit the client code.

> Completions will only work if lowering works on in-progress code

Regarding local completion, I think it's okay to depend on the current on-memory `FileInfo`. Even when editing base functions or similar, we will create `FileInfo` for the base files.

> Add tests

Currently, there's nothing to mock up the whole LS setup, so unit testing `to_completion` is fine for now. However, let's definitely add them.

> Some cases of internal symbols leaking through, e.g. `#fl_parse#0`

If we can figure out JL's internals, we'd like to solve this from the root, but for now, I think a dirty solution like filtering anything starting with `#` is okay.

aviatesk:
@mlechu Sorry for the hasty changes, but to make it easier to develop #20 along with this
PR, I merged #21 first and updated this PR.
For now, I've merged your implementation into `master` almost as it was. I've turned off the local completion feature though, and made a few small changes myself though. You can check them here: https://github.com/aviatesk/JETLS.jl/pull/19/commits/3c1f1eb475f577e673f0a37aa5ed716290759258
Please let me know if you see any issues with that commit.

mlechu:
> I merged https://github.com/aviatesk/JETLS.jl/pull/21 first and updated this PR.

LGTM.  I still have some tests and fixes to push, so I'll push them here when they're done.

> Does this mean that even if we sort the order returned on the server side, it's not reflected on the client side? 

Not yet.  Looked into this a little bit—apparently it's usually the client who is responsible for sorting (maybe to do things like remembering the last choice, which isn't sent to the server).  The server can send `sortText` and `filterText`, but I don't understand what the LSP spec says about those, and they can't be that powerful given that they're just strings.

I assumed in testing that the default vscode client was sorting the completions in some smart way, but it's actually sorting alphabetically...  we should really change it to respect the server's ordering.  

> 
> > Completions will only work if lowering works on in-progress code
> 
> Regarding local completion, I think it's okay to depend on the current on-memory `FileInfo`. Even when editing base functions or similar, we will create `FileInfo` for the base files.

I meant that something like an unterminated `let` block would stop completions from happening.  Thinking about that more now, we should always get a parsing or lowering error to show the user in that case, which isn't so bad.

> If we can figure out JL's internals, we'd like to solve this from the root, but for now, I think a dirty solution like filtering anything starting with `#` is okay.

Will take a look...

src/completions.jl
mlechu: Yup, I think it would make sense to handle `completionItem/resolve` whenever we start caching the results of lowering

test/completions.jl
```diff
@@ -0,0 +1,168 @@
+using Test
+using JETLS: unique, byte_ancestors, greatest_lowerable, is_relevant,
+    cursor_bindings, to_completion, CompletionItem
+
+import JuliaLowering: SyntaxTree, JuliaLowering as JL
+import JuliaSyntax as JS
```
aviatesk:
```suggestion
using JETLS.JuliaLowering: JuliaLowering as JL
using JETLS.JuliaSyntax: JuliaSyntax as JS
```
We need to use relative imports within test files, unless we declare them explicitly in test/Project.toml. I'd recommend using the relative imports since it allows us to use the same versions of JS/JL always.

src/utils.jl
```diff
@@ -66,12 +65,16 @@ function xy_to_offset(fi::FileInfo, pos::Position)
...
 end
+xy_to_offset(fi::FileInfo, pos::Position) = xy_to_offset(fi.parsed_stream.textbuf, pos)
 
 """
 Convert a 1-based byte offset to a 0-based line and character number
 """
-function offset_to_xy(fi::FileInfo, b::Integer)
-    sf = JuliaSyntax.SourceFile(fi.parsed_stream)
+function offset_to_xy(ps::Union{AbstractString, JuliaSyntax.ParseStream}, b::Integer)
+    sf = JuliaSyntax.SourceFile(ps)
```
aviatesk:
```suggestion
    sf = JS.SourceFile(ps)
```
Let's define
```julia
using JuliaSyntax as JS
using JuliaLowering as JL
```
within JETLS.jl and use these aliases consistently across the codebase.

src/completions.jl
```diff
@@ -306,10 +292,5 @@ function handle_CompletionRequest(s::ServerState, msg::CompletionRequest)
                 CompletionList(
                     ; isIncomplete = true,
                     items,
-                    # itemDefaults = (
-                    #     ; data = (
-                    #         ; position = msg.params.position,
-                    #         translated_pos = b,
-                    #         untranslated_pos = offset_to_xy(fi, b)))
```
aviatesk, line 313: Also, can we use this style for calling functions with keyword arguments and making namedtuples?
```julia
itemDefaults = (; 
    data = (; 
        position = msg.params.position,
        translated_pos = b,
        untranslated_pos = offset_to_xy(fi, b)))
```

src/completions.jl
aviatesk: I defined some LSP definitions of implementing `completionItem/resolve` request, and started to use it for retrieving documentations of global symbols lazily.
Now you can specify `data = CompletionData(#=needs_resolve=#true)` for `CompletionItem` generated by `to_complete` to indicate it needs further resolution (handled by `resolve_completion_item`).

We probably need to extend `CompletionData` so that we can reconstruct enough data structures to call `JL.showprov` from serialized JSON data, which can be a bit tricky since JL data structure is quite complicated. We may be able to use some hacks as I did for `state.completion_module` to keep around JL data structures temporarily across a completion request -> completion resolution request round trip.
But this is kind of low priority in terms of finishing this PR, as it looks like calling `showprov` isn't so computationally heavy.

aviatesk:
I found that the implementations of "global" and "local" completion are quite intertwined, so I merged #20 into this branch and would like to continue development based on this PR.
I haven't rebased, so you should be able to just pull the changes. But you need to check out the JET branch at https://github.com/aviatesk/JET.jl/pull/707. At this stage, it might be better to check in Manifest.toml to force the branches for JET and JL.

I feel this PR is getting closer to being ready to merge.
Here are the TODOs I can think of right now:
- [x] 1. Make `local_completions` complete symbols that can not be completed by `global_completions!`
- [x] 2. Finish tests for `local_completions!`
- [x] 3. Add tests for `global_completions!`
- [x] 4. Figure out the VSCode sorting issue

I might need to explain point 1. Basically, I think things where `binding.kind===:global` don't need to be included in the list completed by `local_completions!`. Think, e.g.
```julia
const xxx = 42
function xhello(x::String)
    x#=cursor=#
    return "Hello, $x"
end
```
In a situation like the one above, currently you get completions like this:
Screenshot 2025-05-15 at 20 21 05
I think `xxx` and `xhello` don't need to be completed by `local_completions!` because the one from `global_completions!` is more accurate. This is because global symbols that `local_completions!` can complete are limited to what can be parsed in the currently opened file. But `global_completions!` can complete across files, so let's use that instead.

> I assumed in testing that the default vscode client was sorting the completions in some smart way, but it's actually sorting alphabetically... we should really change it to respect the server's ordering.

Yeah.. the specification is really unhelpful here. I'd like to show the local symbols first, but I'm still looking into how to do that... Letting the client to sort them alphabetically is really silly.

src/completions.jl
mlechu: Sure, that looks better.  I was working around my editor, which indents this like
```julia
itemDefaults = (; 
                data = (; 
                        position = msg.params.position,
                        translated_pos = b,
                        untranslated_pos = offset_to_xy(fi, b)))

```

src/completions.jl
```diff
@@ -296,20 +309,20 @@ function handle_CompletionRequest(s::ServerState, msg::CompletionRequest)
...
-    # local_completions!(items, s, uri, msg.params.position)
-    # global_completions!(items, s, uri, msg.params.position)
+    local_completions!(items, s, uri, msg.params.position)
+    global_completions!(items, s, uri, msg.params.position)
+
+    return s.send(
+        ResponseMessage(;
+            id = msg.id,
+            result = CompletionList(;
+                isIncomplete = true,
```
mlechu: Should we make this false?  I originally made it true to be safe, but I can't think of any situation where it would help us to re-compute, especially since the server doesn't get the completion prefix anyway.

`isIncomplete = true`
```
abc|   # -> textDocuent/completion
abcd|  # -> textDocuent/completion
```

`isIncomplete = false`
```
abc|   # -> textDocuent/completion
abcd|  # -> client filters the results from the last request
```

aviatesk: Yeah, I agree. We may need to revisit this in the future when we support property completion by adding `.` to `triggerCharacters`, but for now, I think it's fine to make it `false`.

test/completions2.jl
```diff
@@ -0,0 +1,58 @@
+module completions2
```
aviatesk: Just for the future reference: I added this test to exercise the whole completion pipeline but it feels like this is far too complex, and it’s exactly the kind of thing that would likely break in the future e.g. when LS were made async, so I feel it could become quickly technical debt down the road. 

For now, I think I will have this test, but it might be better to factor out the part that creates the file cache so we can easily write proper unit tests.

src/completions.jl
mlechu: I've changed it, but feel free to change it back for properties or if it messes with sorting behaviour

src/completions.jl
```diff
@@ -273,43 +269,75 @@ function find_file_module(state::ServerState, uri::URI, pos::Position)
...
+    global_completions!(items, s, uri, msg.params.position)
+
+    # show local completions first, and then global completions
+    fnames = fieldnames(CompletionItem)
+    for idx in eachindex(items)
+        # Hack to set fields on an immutable CompletionItem...
+        mutable_ci = Dict(zip(fnames, [getfield(items[idx], f) for f in fnames]))
+        mutable_ci[:sortText] = lpad(string(idx), SORTTEXT_WIDTH, '0')
+        items[idx] = CompletionItem([get(mutable_ci, f, nothing) for f in fnames]...)
+    end
```
aviatesk: I think we can just specify `sortText = "00000000"` for the items collected by `local_completions!` and `sortText = "00000001"` for those by `global_completions!`, and let VSCode to sort those within a group with the same `sortText`.

src/completions.jl
```diff
@@ -79,6 +76,11 @@ function is_relevant(ctx::JL.AbstractLoweringContext,
                      binding::JL.BindingInfo,
                      cursor::Int)
     !binding.is_internal &&
+
+        # TODO: Temporary restriction to prevent duplicate completions
+        # This is too strict; we miss locally-declared globals.
+        binding.kind !== :global &&
```
aviatesk: JET can find any symbols declared with `global` no matter where it's declared.
Screenshot 2025-05-16 at 17 49 55

Currently it may wrongly recognize module context due to a bad implementation of https://github.com/aviatesk/JET.jl/pull/707 thus it needs `nothing` at the end of `module M` though (will be fixed along with https://github.com/aviatesk/JET.jl/issues/709).

src/completions.jl
mlechu: Maybe, although that also rules out sorting within the locals or globals.

aviatesk: Right, at the moment, AFAIU we are not sorting within locals or globals, but in the future we may sort local completions especially by distance from the cursor position. I implemented my approach for the sake of code simplicity, but if you want to implement a proper sorting order, it's just fine to revert it. If possible, I'd prefer to specify it when constructing the `CompletionItem` though.

aviatesk:
Thanks for making the great progress on this PR @mlechu. Confirmed this PR now passes all the tests added, so let's merge it.

src/completions.jl
aviatesk: EDIT: `global_completions!` (which is based on JET) currently can't return symbols declared within function scope only, so we still need `local_completions!` for filling in those symbols. Maybe further tweak on JET's internal allow us to fully rely on `global_completions!` for collecting all global symbols available in the module context at the cursor.

aviatesk (approved):

PR #20, opened 2025-05-13
implement global completions

aviatesk:
Built on top of https://github.com/aviatesk/JETLS.jl/pull/19, requires https://github.com/aviatesk/JET.jl/pull/707.

To help with future collaboration, first, we will likely extract the basic infrastructure part from https://github.com/aviatesk/JETLS.jl/pull/19 and merge it in a separate PR. Then, we'll rebase both PRs so we can implement local/global independently.


aviatesk:
Merged into #19.

PR #23, opened 2025-05-19
Small completion fixes

mlechu:
- Sort locals by distance from the cursor
- HACK: We can't expand macros through JuliaLowering yet---even if we grab the
  module context, JuliaLowering needs to see the macro definition.  To fix local
  completions not working when macrocalls are around the cursor, trim macrocalls
  from the tree before lowering.
- Remove code attempting to find LambdaBindingInfo for completions (I was
  mistakenly searching our pre-scope-resolution data structures, so the result
  was always nothing).  This was just to give locals a "function" type if we can
  see that they're called, so probably not worth maintaining that much code for.
- Move the `showprov` on local completions from ci.detail to ci.documentation
  (after figuring out where different clients put these in the UI).

Also add a snippet to the setup instructions for the emacs LSP client

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,19 @@ A new language server for the Julia programming language. In a nascent stage of 
 - Press `▷` to run the launch configuration (<kbd>F5</kbd>).
 - In the [Extension Development Host](https://code.visualstudio.com/api/get-started/your-first-extension#:~:text=Then%2C%20inside%20the%20editor%2C%20press%20F5.%20This%20will%20compile%20and%20run%20the%20extension%20in%20a%20new%20Extension%20Development%20Host%20window.) instance of VSCode, open a Julia file.
 
+#### Other editors
+
+Minimal Emacs (eglot client) setup:
+```lisp
+(add-to-list 'eglot-server-programs
+             '(((julia-mode :language-id "julia")
+                (julia-ts-mode :language-id "julia"))
+               "julia"
+               "--startup-file=no"
+               "--project=/path/to/JETLS.jl"
+               "/path/to/JETLS.jl/runserver.jl"))
+```
+
 ### Notes
 
 In JETLS, since we need to use packages that aren’t yet registered
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -11,10 +11,17 @@ function completion_is(ci::CompletionItem, ckind::Symbol)
         (ci.labelDetails.description === "argument" && ckind === :local)
 end
 
-function get_sort_text(ckind::Symbol, offset::Int)
-    # show non-global completions first in the list
-    n = ckind === :global ? (10000000 + offset) : offset
-    lpad(n, 8, '0')
+let sort_texts = Dict{Int, String}()
+    for i = 1:1000
+        sort_texts[i] = lpad(i, 4, '0')
+    end
+    _, max_sort_text = maximum(sort_texts)
+    global function get_sort_text(ckind::Symbol, offset::Int)
+        if ckind === :global
+            return max_sort_text
+        end
+        return get(sort_texts, offset, max_sort_text)
+    end
 end
 
 # local completions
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -24,6 +24,19 @@ A new language server for the Julia programming language. In a nascent stage of
...
+Minimal Emacs (eglot client) setup:
+```lisp
+(add-to-list 'eglot-server-programs
+             '(((julia-mode :language-id "julia")
+                (julia-ts-mode :language-id "julia"))
+               "julia"
+               "--startup-file=no"
+               "--project=/path/to/JETLS.jl"
+               "/path/to/JETLS.jl/runserver.jl"))
+```
```
aviatesk, line 38: Wow, can Emacs set up a language server this easily? That's amazing.

Considering we might need to add some features on the client side, it's okay to check in the client's .el file into this repository for now (if you want).

src/completions.jl
```diff
@@ -13,6 +11,12 @@ function completion_is(ci::CompletionItem, ckind::Symbol)
         (ci.labelDetails.description === "argument" && ckind === :local)
 end
 
+function get_sort_text(ckind::Symbol, offset::Int)
+    # show non-global completions first in the list
+    n = ckind === :global ? (10000000 + offset) : offset
+    lpad(n, 8, '0')
+end
```
aviatesk: This function includes `lpad` string operation and seems to have a small allocation cost every time it's called. So, maybe an approach with pre-allocation like below would be a bit better for performance. Thinking realistically, sorting 1000 candidates would be enough:
```suggestion
let sort_texts = Dict{Int, String}()
    for i = 1:1000
        sort_texts[i] = lpad(i, 4, '0')
    end
    max_sort_text, _ = maximum(sort_texts)
    global function get_sort_text(ckind::Symbol, offset::Int)
        if ckind === :global
            return max_sort_text
        end
        return get(sort_texts, offset, max_sort_text)
    end
end
```

aviatesk:
> Sort locals by distance from the cursor

:+1:

> HACK: We can't expand macros through JuliaLowering yet---even if we grab the  
    module context, JuliaLowering needs to see the macro definition. To fix local  
    completions not working when macrocalls are around the cursor, trim macrocalls  
    from the tree before lowering.

It's a shame, but it might be difficult to deal with this until we have a way to enable JL fully in the base. As a temporary fix, I think this hack is unavoidable.

README.md
aviatesk: By the way, it's pretty handy for development to have a command to restart the language server: https://github.com/aviatesk/JETLS.jl/blob/e8d79c60feef96de4962fe45cc2095c389dc2a73/extension.ts#L59-L67

runserver.jl [automatically starts Revise](https://github.com/aviatesk/JETLS.jl/blob/e8d79c60feef96de4962fe45cc2095c389dc2a73/runserver.jl#L30), so when you make changes that Revise can handle it should dynamically update the LS behavior. But if we change types or similar things that Revise currently can't handle, we have to restart the LS process itself.
I'm not sure if Emacs has a way to implement something like this though.
Just a comment. It might be faster to restart Emacs itself.

src/completions.jl
mlechu: Thanks!  We could probably also limit the number of completions we send and set `isIncomplete = true`. I don't think anyone is going to scroll through 1000 completions before typing more characters in testing or real use.

README.md
mlechu: > it's okay to check in the client's .el file into this repository for now

Not much to check in (eglot is built in and handles most of the work), so I'm OK leaving this in the readme for now.  Later we could do something slightly fancier like the emacs wrapper package for LanguageServer.jl (https://github.com/non-Jedi/eglot-jl) or just suggest they add JETLS to that.

Not sure about adding client features, but if we end up adding something to the visual studio code extension I can look into writing something equivalent.

> I'm not sure if Emacs has a way to implement something like this though.

M-x eglot-reconnect works in this case

PR #24, opened 2025-05-20
improve error handling and make `local_completions!` more robust

aviatesk:
Extend the `try`/`catch` currently only enabled for completions to all
other handlers so the server won’t crash immediately during development.

In tests, disable the `try`/`catch` block to allow immediate detection
of message handler errors.

Note: Once merged, run `Pkg.instantiate()` again for installing Preferences.jl into JETLS environment.

### DIFF
--- a/test/test_completions.jl
+++ b/test/test_completions.jl
@@ -5,12 +5,15 @@ using JETLS
 using JETLS: JL, JS
 using JETLS: cursor_bindings, to_completion, CompletionItem, completion_is
 
-function get_local_completions(s::String, b::Int)
+function get_cursor_bindings(s::String, b::Int)
     ps = JS.ParseStream(s)
     JS.parse!(ps; rule=:all)
     st0 = JS.build_tree(JL.SyntaxTree, ps)
-    out = cursor_bindings(st0, b)
-    map(o->to_completion(o[1], o[2], o[3]), out)
+    return cursor_bindings(st0, b)
+end
+
+function get_local_completions(s::String, b::Int)
+    return map(o->to_completion(o[1], o[2], o[3]), get_cursor_bindings(s, b))
 end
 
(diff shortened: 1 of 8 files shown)


test/test_completions.jl
```diff
@@ -179,6 +182,23 @@ let code = """
...
+let code = """
+    function fo#=cursor=#
+    """
+    b = first(findfirst("#=cursor=#", code))
+    @test get_cursor_bindings(code, b) === nothing
+end
+let # XXX somehow wrapping within `module A ... end` is necessary to get `xx` completion for this incomplete code
+    s = """
+    module A
+    function foo(xx, y=x#=cursor=#)
```
aviatesk, line 195: I found this test case triggers this logging:
```
┌ Info: Found two bindings with the same name:
│   binfo = JuliaLowering.BindingInfo(27, "xx", :argument, 654, nothing, nothing, 0, false, false, false, true, false, false, false)
└   (bscopeinfos[prev])[1] = JuliaLowering.BindingInfo(25, "xx", :argument, 637, nothing, nothing, 0, false, false, false, true, false, false, false)
```

mlechu: I can wrap this in JETLS_DEV_MODE too; it's mostly informational

mlechu (approved):

test/test_completions.jl
aviatesk: Good.

By the way, do you think we shouldn't have this test case? I think supporting completion for incomplete user code is pretty useful, but from what I understand, it seems to depend quite a bit on JL's internal details. More specifically, I think this test case depends on whether `jl_lower_for_completion` errors, making it potentially fragile against JL internals.

(We should wrap `jl_lower_for_completion` in a `try`/`catch` block anyway though, since expansion of user macros can raise arbitrary errors, although this is more of a future thing).

PR #26, opened 2025-05-26
minor refactoring on `byte_ancestors`

aviatesk:
- Changed the algorithm to mainly work on `byte_range::UnitRange{Int}` rather than checking on the first and last byte of the range.
- Some performance improvements by avoidng allocations
- Make Revise work on this file
- Avoid recursion to be able to handle possibilities of encountering huge AST

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -57,22 +57,23 @@ If we know that parent ranges contain all child ranges, and that siblings don't
 have overlapping ranges (this is not true after lowering, but appear to be true
 after parsing), each tree in the result will be a child of the next.
 """
-function byte_ancestors(st0::JL.SyntaxTree, r::UnitRange{Int})
-    function byte_ancestors_(st::JL.SyntaxTree, l::JL.SyntaxList)
-        (JS.numchildren(st) === 0) && return l
+function byte_ancestors(st0::JL.SyntaxTree, rng::UnitRange{Int})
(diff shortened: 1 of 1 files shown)


aviatesk:
Going to merge this. @mlechu let me know if you have any comments on the new impl of `byte_ancestors`.

src/completions.jl
```diff
@@ -44,31 +47,35 @@ function deduplicate_syntaxlist(sl::JL.SyntaxList)
...
 """
-function byte_ancestors(st::JL.SyntaxTree, b::Int, b2::Int=b)
-    function byte_ancestors_(st::JL.SyntaxTree, l::JL.SyntaxList)
-        (JS.numchildren(st) === 0) && return l
-        cis = findall(ci -> (b in JS.byte_range(ci) && b2 in JS.byte_range(ci)),
-                      JS.children(st))
-        append!(l, map(ci -> st[ci], cis))
-        for c in JS.children(st)
-            var"#self#"(c, l)
+function byte_ancestors(st0::JL.SyntaxTree, rng::UnitRange{Int})
```
mlechu, line 60: LGTM.  This isn't important, but I've been using `st0` as convention for "parsed, unlowered syntax tree" and `byte_ancestors` should work on any syntax tree.

aviatesk: Oh, I see, that was the intention. Good to know. I'll fix it along with other PRs.

PR #27, opened 2025-05-26
Add guards against invalid positions in xy_to_offset

aviatesk:
Prevents errors when language clients send invalid `Position` objects by adding bounds checks and early breaks. Function now gracefully handles out-of-bounds line numbers and character positions.

---

Just for reference:
I particularly found this problem occurring in the Zed editor.
In Zed, such invalid positions are passed especially when completion is triggered after deleting characters, where completion requests with valid and invalid positions seem to be sent in quick succession.
I feel like this is probably a bug on Zed's side, but I think there's no harm in making this function's implementation more robust, so I'm adding this fix.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -49,25 +49,40 @@ get_fileinfo(s::ServerState, t::TextDocumentIdentifier) = get_fileinfo(s, URI(t.
 # These functions do the conversion.
 
 """
-Convert 0-based `(;line = y, character = x)` to a 1-based byte offset
+    xy_to_offset(code::Vector{UInt8}, pos::Position)
+    xy_to_offset(fi::FileInfo, pos::Position)
+
+Convert 0-based `pos::Position` (equivalent to `(; line = y, character = x)`) to a 1-based byte offset.
+Basically, `pos` is expected to be valid with respect to `code`.
+However, some language server clients sometimes send invalid `Position`s,
+so this function is designed not to error on such invalid `Position`s.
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -49,25 +49,40 @@ get_fileinfo(s::ServerState, t::TextDocumentIdentifier) = get_fileinfo(s, URI(t.
...
 """
-Convert 0-based `(;line = y, character = x)` to a 1-based byte offset
+    xy_to_offset(code::Vector{UInt8}, pos::Position)
+    xy_to_offset(fi::FileInfo, pos::Position)
+
+Convert 0-based `pos::Position` (equivalent to `(; line = y, character = x)`) to a 1-based byte offset.
+Basically, `pos` is expected to be valid with respect to `code`.
+However, some language server clients sometimes send invalid `Position`s,
+so this function is designed not to error on such invalid `Position`s.
+Note that the byte offset returned in such cases has almost no meaning.
```
aviatesk, line 59: Additionally, when such invalid positions are detected, it might be good to have `local_completions!` bail out early. What do you think?

mlechu: Thanks for testing on another editor!  The current method looks reasonable.  I'm not sure if we should bail out, but if this bug makes local completions inaccurate often, that would make sense to do.

aviatesk: Yep, I agree. Thanks for reviewing this.

PR #31, opened 2025-05-27
Implement LaTeX symbols and emoji completion

abap34:
Implemented completion for LaTeX symbols and emoji. 

https://github.com/user-attachments/assets/0ee649ae-5404-4553-8b12-fe0858cd60fc

Now, items from `REPL.REPLCompletions.latex_symbols` and `REPL.REPLCompletions.emoji_symbols` appear in the completion candidates.

## Overview of the Implementation

A naive approach of simply setting the value in the `insertText` field of `CompletionItem` does not work as expected. For example, when typing `\alp` and triggering completion, it results in `\α`.
I haven’t checked the client implementation in detail, so I'm not sure, but it's likely that the word boundary recognition is incorrect.  
(See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=Completion%20item%20provides%20an,filtered%20on%20the%20client.) (And to begin with, `\` and `alp` are separate tokens.) 

So instead of relying on the client to interpret the text, the approach of using `TextEdit` to explicitly control the replacement range seems good. This can be done by scanning the token sequence.

## Remaining Tasks

- [ ] Adjust display info: Now simply display the converted characters in the `documentation`, I personally find this the clearest, but there might be better alternatives. It is easy to change, so I'll leave it for the final step.
- [x] Tests

### DIFF
--- a/test/test_completions.jl
+++ b/test/test_completions.jl
@@ -367,6 +367,9 @@ end
     end
 end
 
+# Latex&emoji
+# ===========
+
 function test_backslash_offset(code::String, expected_offset=nothing)
     text, positions = get_text_and_positions(code, r"#=cursor=#")
     @assert length(positions) == 1 "test_backslash_offset requires exactly one cursor marker"
@@ -539,6 +542,58 @@ end
     end
     let code = "angle = \\phi#=cursor=#"
         test_backslash_offset(code, 9)
+    end
+end
+
+@testset "Latex/emoji completion" begin
+    state = JETLS.ServerState(identity)
+    filename = "test_latex_emoji.jl"
+    uri = JETLS.URI(filename)
+
+    # `\`-mark should trigger latex completion
+    let text = """
+        function foo(α, β)
+            \\
+        end
+        """
+        JETLS.cache_file_info!(state, uri, 1, text, filename)
+        params = CompletionParams(;
+            textDocument=TextDocumentIdentifier(string(uri)),
+            position=Position(;line=1,character=5),
+            context=CompletionContext(;
+                triggerKind=CompletionTriggerKind.TriggerCharacter,
+                triggerCharacter="\\"))
+        items = JETLS.get_completion_items(state, uri, params)
+        @test any(items) do item
+            item.label == "\\alpha"
+        end
+        @test !any(items) do item
+            item.label == "foo" || # should not include global completions
+            item.label == "β"      # should not include local completions
+        end
+    end
+
+    let text = """
+        function foo(α, β)
+            \\:
+        end
+        """
+        JETLS.cache_file_info!(state, uri, 2, text, filename)
+        params = CompletionParams(;
+            textDocument=TextDocumentIdentifier(string(uri)),
+            position=Position(;line=1,character=6),
+            context=CompletionContext(;
+                triggerKind=CompletionTriggerKind.TriggerCharacter,
+                triggerCharacter=":"))
+        items = JETLS.get_completion_items(state, uri, params)
+        @test any(items) do item
+            item.label == "\\:pizza:"
+        end
+        @test !any(items) do item
+            item.label == "foo" || # should not include global completions
+            item.label == "α"      # should not include local completions
+            item.label == "β"      # should not include local completions
+        end
     end
 end
 


aviatesk:
Thank you for implementing it so quickly. I think it's excellent progress.

I have a few minor feedback points, but could you rebase first?
It seems there are some conflicts, especially in `triggerCharacters` and similar areas.

src/completions.jl
```diff
@@ -24,6 +24,44 @@ let sort_texts = Dict{Int, String}()
...
+    elseif curr_idx > 1 && tokens[curr_idx - 1].orig_kind == JS.K"\\" 
+        return tokens[curr_idx - 1].next_byte - 1
+    # example 3
+    elseif curr_idx > 2 && tokens[curr_idx - 2].orig_kind == JS.K"\\" && tokens[curr_idx - 1].orig_kind == JS.K":"
+        return tokens[curr_idx - 2].next_byte - 1
+    # example 4, 5
+    else 
+        return nothing
+    end
+end
```
aviatesk: This subroutine looks sensible but probably needs unit tests.

src/completions.jl
```diff
@@ -367,6 +444,7 @@ end
 function get_completion_items(state::ServerState, uri::URI, pos::Position)
     items = Dict{String, CompletionItem}()
     # order matters; see local_completions!
+    add_emoji_latex_completions!(items, state, uri, pos)
```
aviatesk: When we complete these symbols I think we can just bail out early and skip other completions.

src/completions.jl
```diff
@@ -338,6 +376,45 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
...
+                description=description
+            ),
+            kind=CompletionItemKind.Unit,
+            documentation=val,
+            textEdit=TextEdit(;
+                range = Range(;
+                    start = backslash_pos,
+                    var"end" = pos),
+                newText = val),
+            sortText=get_sort_text(0, false),
```
aviatesk: We probably don't need to specify this, if we bail out and skip the other completions.

src/completions.jl
```diff
@@ -338,6 +376,45 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
...
+    fi === nothing && return items
+
+    backslash_offset = get_backslash_offset(state, uri, pos)
+    backslash_offset === nothing && return items
+    backslash_pos = offset_to_xy(fi, backslash_offset)
+
+    function create_ci(key, val, description)
+        return CompletionItem(
+            label=key,
+            labelDetails=CompletionItemLabelDetails(
```
```suggestion
        return CompletionItem(;
            label=key,
            labelDetails=CompletionItemLabelDetails(;
```
Use explicit semicolons.

src/completions.jl
```diff
@@ -338,6 +376,45 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
...
+                description=description
+            ),
+            kind=CompletionItemKind.Unit,
```
```suggestion
            kind=CompletionItemKind.Snippet,
```
`Snippet` seems to better fit here?

aviatesk:
> So instead of relying on the client to interpret the text, the approach of using `TextEdit` to explicitly control the replacement range seems good. This can be done by scanning the token sequence.

This sounds to be a sensible choice. The specification also recommends using `TextEdit` when possible in general, as it allows us to apply completions as expected no matter how clients implement word boundaries.

> * Adjust display info: Now simply display the converted characters in the `documentation`, I personally find this the clearest, but there might be better alternatives. It is easy to change, so I'll leave it for the final step.

The current arrangement seems to be most reasonable and I'm just fine with this.

> * Tests

Testing a LSP feature thoroughly is rather tricky, but it is possible to test using the method implemented in the latter part of test_completions.jl.
However, I feel that the current method of LSP feature testing depends on intricate details of the LS's internal implementation (such as the use of `JETLS.cache_file_info!`), and will likely need to be modified in response to future changes, which I'm not very satisfied with.
For now, it's fine to just perform unit tests on subroutines with relatively clear purposes, like `get_backslash_offset`.

abap34:
Thanks! I did a rebase and the conflicts are resolved now.
I will add the test and fix the points your raised.

aviatesk:
I added several fixes and commits that add tests.
When you have time, it might be good if you could check the added commits for future reference.
There are still some unsatisfactory points, but I'll merge it as is for now.
Thank you for your contribution despite being busy!

Remaining TODOs for the future:
- [x] Emoji completion doesn't trigger when typing `\:|`
- [x] Same issue when typing `\:+|`
- [x] For cases like `\alp|`, would like to improve `sortText` so that `\alpha` appears first

PR #34, opened 2025-05-29
Implement signature help

mlechu:
Shows method signatures when the user is typing out a function call.  

## Examples
| Emacs | Visual Studio Code |
| ----------- | ----------- |
|emacs1|vscode1|
|emacs2|vscode2|
|emacs3|vscode3|
|emacs4|vscode4|

## Implemented
- Parameter highlighting for positional and keyword arguments
- Filtering of signatures (no type info used yet, just by number of positional arguments and names of keyword arguments)
- Support for keyword args before the semicolon in a call
- Tolerance of some partially-incorrect syntax, e.g. `foo(|`, `foo(1,,|)`

## Todo
- `dotcall` support (simple to implement, I just need to decide what would go in the UI)

## One day
- Provenance would be nice here to get the original signature from source instead of the output of `show(::Method)`

### Misc griping
Ran into lots of flakiness in the clients.  I estimate half of these are my own bugs and I've yet to figure it out, but I figured I should write them down somewhere.
  - emacs: Specifying param labels using ranges is off by one (see above)
  - emacs: Highlighting params doesn't work on the first signature (see above)
    - This works with other language servers, so worth investigating
  - vscode: The signature popup closes when typing `.` or `;` (and others)
    - We just need to add everything to the trigger character list, I guess
  - protocol (vscode): You can't not highlight your parameters.  From the LSP docs:
    ```
	/**
	 * The active parameter of the active signature. If omitted or the value
	 * lies outside the range of `signatures[activeSignature].parameters`
	 * defaults to 0 if the active signature has parameters. If
	 * the active signature has no parameters it is ignored.
	 * In future version of the protocol this property might become
	 * mandatory to better express the active parameter if the
	 * active signature does have any.
	 */
    ```
    but it doesn't always make sense to highlight a parameter, e.g. `f(1,2,a...|)`.  Eglot does the reasonable thing in my opinion, and vscode will fall back to highlighting the first.
  - vscode: It is possible to break the param highlighter completely if it doesn't like the characters in a string parameter label, which at least fixes the issue above


src/LSP/language-features/signature-help.jl
```diff
@@ -0,0 +1,251 @@
...
+    """
+    The client supports the following `SignatureInformation`
+    specific properties.
+    """
+    signatureInformation::Union{Nothing, @interface begin
+        """
+        Client supports the follow content formats for the documentation
+        property. The order describes the preferred format of the client.
+        """
+        documentationFormat::Union{Nothing, Vector{MarkupKind}} = nothing
```
aviatesk:
```suggestion
        documentationFormat::Union{Nothing, Vector{MarkupKind.Ty}} = nothing
```
This is a very confusing and unfortunate aspect of `@namespace`/`@interface`, but the `A` introduced by `@namespace A::T` is actually a module, and the type of enum data "A" is accessed via `A.Ty`. I once considered whether there was a solution that would eliminate the need for `.Ty`, but it's not possible to directly convert TS `namespace` to Julia equivalent, so we've settled on the current situation.

mlechu:
> I apologize for the delay in the review

No worries, review whenever you have the time.

> I believe the approach taken in this PR is based on JL, but could you explain why a JL-based implementation is necessary

I do use JuliaLowering's syntax tree structure for parsed code, but nothing is lowered, and we could switch to `JS.SyntaxNode` if we wanted.  I used `JL.SyntaxTree` for consistency with completions, and also because [some comments](https://github.com/c42f/JuliaLowering.jl/tree/4b12ab19dad40c64767558be0a8a338eb4cc9172?tab=readme-ov-file#syntax-trees) in the JuliaLowering readme made me think `JL.SyntaxTree` is meant to be an updated version of `JS.SyntaxNode`.

> rather than the approach being attempted in [JuliaLang/julia#57767](https://github.com/JuliaLang/julia/pull/57767) (an advanced JS-based approach with setting up a custom node)?

We don't use custom nodes currently, but `SyntaxTree` would actually make an approach like this much easier, because we could add new attributes (e.g. `:parent`, `:cursor_offset`) to any subset of nodes by doing a pass over an existing tree instead of defining a new type of node and translating.

Our current solution to finding the nodes containing the cursor is `byte_ancestors`, which is messier, but flexible, and works well after lowering when byte ranges of parents don't necessarily contain their children.

> Are there cases where scope resolution using JL is advantageous for implementing this feature?

No local scope resolution happens right now (our signatures are just `methods()` from the cursor's current module) but lowering would be necessary if we were to try.

src/signature-help.jl
```diff
@@ -0,0 +1,325 @@
...
+
+"""
+textDocument/signatureHelp is requested when one of the negotiated trigger
+characters is typed.  Eglot (emacs) requests it more frequently.
+"""
+function handle_SignatureHelpRequest(server::Server, msg::SignatureHelpRequest)
+    state = server.state
+    uri = URI(msg.params.textDocument.uri)
+    fi = get_fileinfo(state, uri)
+    mod = find_file_module!(state, uri, msg.params.position)
```
aviatesk: Since `find_file_module!` dynamically changes the global state `state.completion_module` based on `pos`,
it is safer to use `find_file_module`. `state.completion_module` is an internal state only for supporting `completionItem/resolve`, so other features should not modify/use it.

aviatesk:
Since I made quite large-scale changes to `master`, I have added the updates here myself.

src/signature-help.jl
aviatesk: Addressed in [d1f3b5e](https://github.com/aviatesk/JETLS.jl/pull/34/commits/d1f3b5eebe85d06da07533a9e3d950317defa0f1) along with the other changes.

src/signature-help.jl
```diff
@@ -0,0 +1,327 @@
...
+        !isdefined(mod, s) && return nothing
+        return getproperty(mod, s)
+    elseif kind(st0) === K"."
+        @assert JS.numchildren(st0) === 2
+        lhs = resolve_property(mod, st0[1])
+        return resolve_property(lhs, st0[2])
+    end
+    JETLS_DEV_MODE && @info "resolve_property couldn't handle form:" mod st0
+    return nothing
+end
```
aviatesk: As you mention in the docstring, the better approach to do this might be to do something like
https://github.com/JuliaLang/julia/blob/1ea959a8b12f06391c305d600fb6942cb25905f6/stdlib/REPL/src/REPLCompletions.jl#L640-L674

It should allow us to handle cases like `(print("hi"); Base).print`.
Additionally, it might also allow us to account for argument types to some extent (by inferring the entire method call expression, and inspect types of IR elements corresponding to arguments).

However since it's a somewhat advanced approach, it's fine to tackle it in a separate PR. Maybe I'm better fit to write `ConcreteEvaluator` at least.

Also, since it's questionable whether that approach can handle incomplete code, it might be necessary to keep this `resolve_property` as a fallback, but I'm somewhat reluctant to provide two routines to do the same thing.

src/signature-help.jl
```diff
@@ -0,0 +1,327 @@
...
+Resolve a name's value given a root module and an expression like `M1.M2.M3.f`,
+which parses to `(. (. (. M1 M2) M3) f)`.  If we hit something undefined, return
+nothing.  This doesn't support some cases, e.g. `(print("hi"); Base).print`
+"""
+function resolve_property(mod::Module, st0::JL.SyntaxTree)
+    if JS.is_leaf(st0)
+        # Would otherwise throw an unhelpful error.  Is this true of all leaf nodes?
+        @assert JL.hasattr(st0, :name_val)
+        s = Symbol(st0.name_val)
+        !isdefined(mod, s) && return nothing
```
aviatesk:
```suggestion
        !(@invokelatest isdefinedglobal(mod, s)) && return nothing
```

src/signature-help.jl
```diff
@@ -0,0 +1,327 @@
...
+        return getproperty(mod, s)
```
aviatesk:
```suggestion
        return @invokelatest getglobal(mod, s)
```

src/signature-help.jl
```diff
@@ -0,0 +1,327 @@
...
+"""
+function flatten_args(call::JL.SyntaxTree)
+    if kind(call) === K"where"
+        return flatten_args(call[1])
+    end
+    @assert kind(call) === K"call" || kind(call) === K"dotcall"
+    usable = (arg::JL.SyntaxTree) -> kind(arg) != K"error"
+    orig = filter(usable, JS.children(call)[2:end])
+
+    args = JL.SyntaxTree[]
```
aviatesk: Since `JL.SyntaxTree` is an abstract type, doing this should make it type-stable.
```suggestion
    args = typeof(orig)[]
```

src/signature-help.jl
```diff
@@ -0,0 +1,354 @@
...
+        mstr[1:lb], string(strip(mstr[lb+1:end]))
+    end
+
+    # We could show the full docs, but there isn't(?) a way to separate by
+    # method (or resolve items lazily like completions), so we would be sending
+    # many copies.  The user may have seen this already in the completions UI,
+    # too.
+    # documentation = MarkupContent(;
+    #     kind = MarkupKind.Markdown,
+    #     value = string(Base.Docs.doc(Base.Docs.Binding(m.var"module", m.name))))
```
aviatesk: There is a way for doing this, but I agree with that the signature help don't need to show it.

src/signature-help.jl
```diff
@@ -0,0 +1,354 @@
...
+        elseif i in keys(ca.pos_map)
+            lb, ub = get(ca.pos_map, i, (1, nothing))
+            if !isnothing(maybe_var_params) && lb >= maybe_var_params
+                maybe_var_params
+            else
+                lb === ub ? lb : nothing
+            end
+        elseif kind(ca.args[i]) === K"..."
+            # splat after semicolon
+            maybe_var_kwp
```
aviatesk, line 263: I think this code path doesn't handle cases like
```
func(x, xs...) = nothing
func(nothing, xs...|)
```
?

aviatesk:
> I do use JuliaLowering's syntax tree structure for parsed code, but nothing is lowered, and we could switch to `JS.SyntaxNode` if we wanted. I used `JL.SyntaxTree` for consistency with completions, and also because [some comments](https://github.com/c42f/JuliaLowering.jl/tree/4b12ab19dad40c64767558be0a8a338eb4cc9172?tab=readme-ov-file#syntax-trees) in the JuliaLowering readme made me think `JL.SyntaxTree` is meant to be an updated version of `JS.SyntaxNode`.

Okay, got it. From you comment now I think `SyntaxTree` can essentially do the same things as JS+custom node. I re-read the code, and I think most of it can be done with `SyntaxTree`, but since we can use JL, let's use it. Also, if we extend this feature in the future to include type information, using JL will be essential.

src/signature-help.jl
```diff
@@ -0,0 +1,354 @@
...
+
+    mstr = sprint(show, m)
+    mnode = JS.parsestmt(JL.SyntaxTree, mstr; ignore_errors=true)[1]
+
+    params, kwp_i = flatten_args(mnode)
+    has_var_params = kwp_i > 1 && kind(params[kwp_i - 1]) === K"..."
+    has_var_kwp = kwp_i <= length(params) && kind(params[end]) === K"..."
+
+    kwp_map = find_kws(params, kwp_i; sig=true)
+
```
aviatesk, line 182: I feel that parsing the output of `sprint(show, m::Method)` and analyzing candidate methods based on it seems a bit hacky.
Is there a reason for taking this approach instead of using `Base.arg_decl_parts`/`Base.kwarg_decl` which are internally used by `Base.show_method`?
https://github.com/JuliaLang/julia/blob/9108dd08a572e394854d85aa0b2b680cc6a591c3/base/methodshow.jl#L217-L259

src/signature-help.jl
```diff
@@ -0,0 +1,354 @@
+using .JS
+using .JL
+
+# initialization
+# ==============
+
+signature_help_options() = SignatureHelpOptions(;
+    triggerCharacters = ["(", ",", ";"],
```
aviatesk:
```suggestion
    triggerCharacters = ["(", ",", ";", "="],
```
Maybe we should add `=` too as it allows us to get sighelp on cases like
image
?

aviatesk:
Quick question, is it possible to show signature help in this case? Specifically, when the text typed after `;` matches one of the kwargs. Right now, signature help stops here.
Screenshot 2025-06-04 at 00 03 18

src/signature-help.jl
aviatesk: Or, maybe we should register "all" characters (is that overkill?). We could consider cases where a user cancels signature help midway and then starts typing again, and it's a bit annoying that signature help is cancelled the moment `"` is typed right now. I feel like performance tuning `handle_SignatureHelpRequest` might be necessary though.

aviatesk: By the way, as a hack that is only possible with VSCode, when experimenting with various `triggerCharacters`, the hack explained at
https://github.com/aviatesk/JETLS.jl?tab=readme-ov-file#dynamic-registration
might be useful.

src/signature-help.jl
mlechu: This is a bug.  `func(nothing, nothing, xs...|)` works (and is tested).  Since `xs...` is positional, we should hit `elseif i in keys(ca.pos_map)`.

src/signature-help.jl
mlechu: I think it would make sense to put `=` (and maybe more) in `retriggerCharacters` instead.  

Agree the default vscode behaviour is annoying...for reference, eglot (in addition to trigger chars) just sends a signature help request whenever the cursor stops moving for a second or so.  What does Zed do?

> hack that is only possible with VSCode

Cool, will take a look

aviatesk: > I think it would make sense to put `=` (and maybe more) in `retriggerCharacters` instead.

The reason I think `triggerCharacters` might be better than `retriggerCharacters` is that currently, as per https://github.com/aviatesk/JETLS.jl/pull/34#issuecomment-2935837838, when typing up to `func(arg1, arg2; k|)`, the signature help gets canceled, so unless we trigger fresh signature help, it might not be retriggered.
At least in VSCode, once signature help disappears, it doesn't seem to be retriggered. However, eglot might handle this case better.

So, if https://github.com/aviatesk/JETLS.jl/pull/34#issuecomment-2935837838 can be resolved, it might be fine to include `=` in `retriggerCharacter`. But registering it into `triggerCharacters` seems to be more robust and inefficient solution.

> What does Zed do?

The situation with Zed is even worse...
The implementation of signature help in Zed is very primitive, and it doesn't seem to respect `triggerCharacters` at all. I've submitted this as an issue.
https://github.com/zed-industries/zed/issues/31846

mlechu: > if https://github.com/aviatesk/JETLS.jl/pull/34#issuecomment-2935837838 can be resolved

Yup, I'm going to try and exclude `k` from the filtering logic here, as it doesn't make sense to filter on partially-typed kwargs.  I'll test if `=` works well in retriggerCharacters then.

> The situation with Zed is even worse

nooooo.  Thanks for filing the issue haha

src/signature-help.jl
aviatesk: Umm, based on the CI, it looks like this approach isn't working. I don't feel super strongly about it, so we can just revert to `JL.SyntaxTree[]` for now.

src/signature-help.jl
mlechu: I ended up liking `=` in triggerCharacters better, since typing `=` after a misspelled keyword would otherwise close the signature help for good

src/signature-help.jl
mlechu: (data race with the comment below) Discussed in meeting; definitely hacky, but since we want the provenance of parameters in the string, this might be the approach with the least amount of new code.  Changing this later wouldn't be hard, just tedious (we'd copy the body of `show(::Method)` here and either use JuliaSyntax on that, or join strings together while keeping a counter)

aviatesk: As we discussed in the meeting, since we're using the output of `mstr = sprint(show, m::Method)` as `SignatureHelpInformation.label`, we need position information within `mstr` for parameter highlighting, which is why the current implementation is designed this way.

For future improvements, I think we could consider the following options:
- Adjust `Base.show_method` to add an option that removes the ` @ /path/to/method` part from the output (for making the parsing part safe)
- For `Method` objects that JETLS analyzes, maintain the original JS/JL definitions and use them for signature help as well

Anyway, I think we can proceed with this PR as it is for now.

src/signature-help.jl
aviatesk: This enhancement will be tackled in another PR.

src/signature-help.jl
aviatesk: Confirmed this fixed on the latest commit.

src/signature-help.jl
aviatesk: Yeah, I agree that that is a sensible choice.

PR #40, opened 2025-05-31
fix completion request handling for empty files

abap34:
When the user explicitly triggers completion, it's possible that the program is empty. In such cases, attempts to retrieve the current node may fail, so this edge case needs to be handled properly.

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -398,8 +398,9 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
         # check if the cursor is within the macro name context `@xxx|`
         curbyte = xy_to_offset(fi, pos)
         sn = JS.build_tree(JS.SyntaxNode, fi.parsed_stream)
-        if !isempty(sn.children) 
-            curnode = first(byte_ancestors(sn, curbyte-1))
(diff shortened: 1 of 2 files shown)


src/completions.jl
```diff
@@ -398,8 +398,10 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
         # check if the cursor is within the macro name context `@xxx|`
         curbyte = xy_to_offset(fi, pos)
         sn = JS.build_tree(JS.SyntaxNode, fi.parsed_stream)
-        curnode = first(byte_ancestors(sn, curbyte-1))
-        is_macro_invoke = JS.kind(curnode) === K"MacroName"
+        if !isempty(sn.children) 
+            curnode = first(byte_ancestors(sn, curbyte-1))
+            is_macro_invoke = JS.kind(curnode) === K"MacroName"
+        end
```
```suggestion
        ancestors = byte_ancestors(sn, curbyte-1)
        if !isempty(ancestors) 
            curnode = first(ancestors)
            is_macro_invoke = JS.kind(curnode) === K"MacroName"
        end
```

aviatesk (approved):

PR #46, opened 2025-06-02
test: add tests for runserver.jl script

aviatesk:
Add simple tests for the runserver.jl script.
These tests are somewhat hacky but also added timeouts to try to ensure that no dead processes remain when tests finish, even in a case when problems occur during execution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,13 +17,13 @@ jobs:
           # - version: '1' # current stable
           #   os: ubuntu-latest
           #   arch: x64
-          - version: '1.12-nightly' # minimum version supported
+          - version: "1.12-nightly" # minimum version supported
             os: ubuntu-latest
             arch: x64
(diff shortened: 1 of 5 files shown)


aviatesk:
test_runserver.jl works locally, but for some reason it doesn't work on CI, so I've decided to comment out the CI..

Specifically, `bytesavailable(io)` in this part is always `0`:
https://github.com/aviatesk/JETLS.jl/blob/0c299fa63788e1e640a8238eda8a2337e99aefbe/test/test_runserver.jl#L73

And then it hits the timeout here:
https://github.com/aviatesk/JETLS.jl/blob/0c299fa63788e1e640a8238eda8a2337e99aefbe/test/test_runserver.jl#L111

I don't know why this issue occurs only on CI, but it might be a problem specific to `Base.BufferStream`.
However, since there are many other things to prioritize, I'll merge this as is.

PR #52, opened 2025-06-04
LSP: automatically convert `::[DocumentUri|URI]` fields to `URIs2.URI`

aviatesk:
According to the LSP, any fields declared as `DocumentUri` and `URI` types (in the TypeScript definition) are:
> Over the wire, it will still be transferred as a string, but this guarantees that
> the contents of that string can be parsed as a valid URI.

In actual language server implementations, the values of such `DocumentUri` and `URI` type fields are parsed and used as data structures representing URIs that comply with https://datatracker.ietf.org/doc/html/rfc3986,
having `scheme`, `authority`, `path`, `query`, and `fragment` components.

In JETLS, `URIs2.URI` corresponds to such a data structure.

If we were to strictly adhere to the LSP definition, these fields should be defined as `String` type, and `DocumentUri` and `URI` should be aliases to `String`. However, in our Julia version of the LSP definition, for implementation simplicity, we automatically `convert` URI-representing fields to `URIs2.URI` during JSON3 parsing, allowing the language server to directly handle `URIs2.URI`.

---

/cc @pfitzseb Can I hear your thought on this?

### DIFF
--- a/src/JETLS.jl
+++ b/src/JETLS.jl
@@ -278,12 +278,12 @@ function handle_InitializeRequest(server::Server, msg::InitializeRequest)
     workspaceFolders = params.workspaceFolders
     if workspaceFolders !== nothing
         for workspaceFolder in workspaceFolders
-            push!(state.workspaceFolders, URI(workspaceFolder.uri))
+            push!(state.workspaceFolders, workspaceFolder.uri)
         end
     else
         rootUri = params.rootUri
         if rootUri !== nothing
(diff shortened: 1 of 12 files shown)


aviatesk:
From the perspective that LSP.jl is trying to define LSP accurately in Julia, I think this PR is somewhat debatable, but since it actually makes the implementation much simpler, I'd like to merge it for now.
If you have any opinions, please let me know even after the merge.

PR #59, opened 2025-06-06
LSP: add src/LSP/README.md

aviatesk:
Added documentation explaining how the original LSP specification is converted to Julia code. When this document is provided to an LLM along with sections of the LSP spec, I confirmed it produces high-quality conversions.

The original LSP spec markdown can be copied from: https://publish.obsidian.md/jetls/work/JETLS/Language+Server+Protocol+Specification+-+3.17+(raw)


mlechu (approved):
Change looks good; CI failures are not related.  Might be worth renaming LSP/README.md to LSP/AGENTS.md if most of it is for robots to read.

aviatesk:
> Might be worth renaming LSP/README.md to LSP/AGENTS.md if most of it is for robots to read.

Basically, that's the intention, but considering the possibility of making LSP.jl an external package in the future, I've added documentation on how these macros work just in case, so it might be useful for some people. I'll leave it as README.md for now.

PR #61, opened 2025-06-07
Implement Go to Definition for method

abap34:
This PR implements a prototype of Go to Definition for methods.

go_to_def_demo

Currently, it simply displays the all of results from`methods`.

There are several known limitations:

#### 1. It does not work correctly if the function is later replaced:

```julia
f(x) = 1   # Def 1
y = f(1)   # In this implementation, this `f` will navigate to Def2.

f(x) = 2   # Def 2
z = f(1)
``` 

####  2. It does not distinguish between struct field access and module access.

```julia
struct S
  f::Function
end

s = S(x -> x + 1)
s.f(2)  # We need to distinguish whether `s` is a module or some other object.
```

####  3. It does not retrieve the full range of the method definition (only the starting line).

This is expected to be resolved in the future as the analysis infrastructure improves.

#### Remaining tasks in other PRs
-  Support definitions other than methods
-  Handle the binding location properly
-  Type-aware filtering of method candidates

@aviatesk  For now, I think it would be useful to merge this level of prototype implementation.  What do you think? If it looks good, I’ll  add tests 🐩.


aviatesk:
Excellent progress!
Thank you for mentioning the limitations as well.
I'll note it in the Obsidian note, so let's refer back to it as needed and aim to resolve it ultimately.

> For now, I think it would be useful to merge this level of prototype implementation. What do you think? If it looks good, I’ll add tests 🐩.

Yes, I agree.
Let's add some simple tests and start with the easy parts.

src/JETLS.jl
```diff
@@ -395,6 +409,18 @@ function handle_InitializedNotification(server::Server)
...
+        :textDocument, :definition, :dynamicRegistration) === true
+        push!(registrations, definition_registration())
+        if JETLS_DEV_MODE
+            @info "Dynamically registering 'textDocument/definition' upon `InitializedNotification`"
+        end
+    else
+        # NOTE If definition's `dynamicRegistration` is not supported,
+        # it needs to be registered along with initialization in the `InitializeResponse`,
+        # since `DefinitionRegistrationOptions` does not extend `StaticRegistrationOptions`.
+    end
```
aviatesk, line 422: 👍

abap34:
Thank you! Now I will work on tests, etc. (I put an extension in my editor to remove trailing spaces 😅)

aviatesk:
> I put an extension in my editor to remove trailing spaces 😅

The Helix editor seems to have the builtin setting for that (`trim-trailing-whitespace`)?
https://docs.helix-editor.com/master/editor.html

abap34:
Thanks!  It seems it was an option that was supported in the newer version, so I upgraded the helix version and it worked fine! By the way, I added some tests.

EDIT: Fixed one wrong indentation.

test/test_definition.jl
```diff
@@ -0,0 +1,207 @@
...
+    while true
+        pos = findfirst(cursor, code)
+        if pos === nothing
+            break
+        end
+        push!(offsets, pos.start)
+        code = replace(code, cursor => "", count=1)
+    end
+    return offsets, replace(code, cursor => "")
+end
```
aviatesk: Maybe best to use `JETLS.get_text_and_positions`. It handles cases where multiple `matcher`s are found in the given `text`, although your test cases don't need that.

src/definition.jl
```diff
@@ -0,0 +1,117 @@
...
+
+    for i in 2:length(bas)
+        if kind(bas[i]) !== K"."
+            return bas[i - 1]
+        end
+    end
+
+    # Unreachable: we always have toplevel node
+    return nothing
+end
```
aviatesk: Can you elaborate what "best" does mean actually? Also, can't we use `resolve_property` directly?

abap34: In typical non-modal editors like VSCode, the cursor is shown between characters, so in cases like the following, it’s often ambiguous which token is actually “under” the cursor:

```
func│(x)
```

In the current implementation, such cases don’t navigate anywhere. (The cursor is assumed to be in the left bracket.) However, as noted in the comments, rust-analyzer resolves this ambiguity by assigning a priority to each token. (In this case, navigate to `func`)

So “best” here refers to what the user is trying to navigate the most **around** the cursor.
I'll modify the comment to make this clearer.

test/test_definition.jl
abap34: I’ve updated the code to use this: 825554b

src/definition.jl
abap34: updated: 947ec87

aviatesk: I see, Thanks for the explanation.
Regarding the name, I think it's taken from that of rust-analyzer, but still `get_best_node` is indeed somewhat ambiguous, how about `select_navigation_node` (or `select_target_node` like `get_target_node` in the test)?

test/test_definition.jl
```diff
@@ -0,0 +1,195 @@
...
+
+@testset "method_definition_range" begin
+    linenum = @__LINE__; method_for_test_method_definition_range() = 1
+    @assert length(methods(method_for_test_method_definition_range)) == 1
+
+    test_method = first(methods(method_for_test_method_definition_range))
+    method_range = method_definition_range(test_method)
+
+    @test method_range isa JETLS.Location
+    @test endswith(method_range.uri.path, PROGRAM_FILE)
```
aviatesk:
```suggestion
    @test JETLS.URIs2.uri2filepath(method_range.uri) == @__FILE__
```

test/test_definition.jl
```diff
@@ -0,0 +1,195 @@
...
+        (result, uri) ->
+            (length(result) >= 1) &&
+            (any(result) do candidate
+                candidate.uri.path == sin_cand_file &&
+                candidate.range.start.line == (sin_cand_line - 1)
+            end),
+
+        # Co│re.Compiler.tmeet
+        (result, uri) ->
+            (length(result) >= 1),
```
aviatesk: This seems to relate to the implementation of `get_best_node`, but I think in cases like this, it would be more natural to jump to where `Core` is defined rather than jumping to `tmeet`, while I agree with jumping to `tmeet` given cases like `Core.Compiler.tme|et`. (we don't need to implement jumping to modules in this PR though)

src/definition.jl
abap34: That makes sense. Since they seem to use it in a more general context, I think `select_target_node`, which better fits our use case, is a good choice. I’ve pushed that renaming.

aviatesk (commented):
I think this PR is almost close to completion, thanks for making a great progress.

test/test_definition.jl
abap34: That behavior feels more natural to me as well. I’ve updated `select_target_node` so that it no longer follows child module. 432249c

aviatesk (approved):

PR #67, opened 2025-06-10
allow goto definition for general callables

aviatesk:
Although I found several test cases are failing (previously we actually didn't `@test` them).
@abap34 Can you look into the failures and fix them?

### DIFF
--- a/test/test_definition.jl
+++ b/test/test_definition.jl
@@ -2,9 +2,8 @@ module test_definition
 
 using Test
 using JETLS
-using JETLS: JL, JS, select_target_node, method_definition_range, definition_locations,
+using JETLS: JL, JS, select_target_node, method_definition_range,
              get_text_and_positions, xy_to_offset
-
 
 function get_target_node(code::AbstractString, pos::Int)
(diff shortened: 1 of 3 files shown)


test/test_definition.jl
```diff
@@ -183,13 +196,14 @@ include("setup.jl")
             @test raw_res.params.uri == uri
 
             for (pos, tester) in zip(positions, testers)
-                let id = id_counter[] += 1
+                @testset let loc = functionloc(only(methods(tester))),
+                             id = id_counter[] += 1
```
aviatesk: Just a note: `@testset let` is useful, letting us get some context information about which test case is actually failing.

aviatesk:
@abap34 You can add commits directly on this branch rather than making a separate branch on top of this. You can of course cut a new branch if you want.

aviatesk:
Since this PR is a prerequisite for aviatesk/JETLS.jl#70, I will mark the broken test cases as `broken=true` and move forward for now.
Please fix the broken test cases later.

PR #69, opened 2025-06-10
fix filtering ambiguity in LaTeX and emoji completion

abap34:
Some clients (e.g., VSCode) seem to discard completion items when filterText is not explicitly set and the beginning of the label does not match the current input.

This change explicitly sets `filterText`, and I have confirmed that it works both in VSCode and in editors where it previously worked.


aviatesk (approved):

aviatesk:
Great, thanks!

PR #70, opened 2025-06-10
add `resolve_type` functionality for type resolution

aviatesk:
Implement `resolve_type` function that resolves types of syntax nodes in analyzed contexts, which is supposed to replace usages of `resolve_property`:

>     resolve_type(analyzer::LSAnalyzer, context_module::Module, s0::Union{JS.SyntaxNode,JL.SyntaxTree})
>     resolve_type(analyzer::LSAnalyzer, context_module::Module, ex::Expr)
>
> Resolves the type of `s0` in the `context_module` that has been analyzed by the `analyzer`.
> When resolution is successful, it returns the type of the `node` in the extended lattice (e.g., `Int`, `Const`),
> and when resolution fails, it returns `nothing`.
> Note that `s0` is just a raw `SyntaxTree` or `SyntaxNode` that has not been lowered.
>
> This utility has a critical flaw: it cannot correctly handle bindings that are shadowed in local scopes.
> That is, `s0` is always resolved in the global scope of `context_module`.
> Therefore, it may return incorrect results in cases like:
> ```julia
> function foo(x)
>     sin = cos
>     y = sin(x) # resolve_type(analyzer, context_module, :(sin)) would return `Const(sin)` instead of `Const(cos)`
>     return y
> end
> ```
>
> For this reason, significant changes are expected in the near future.
> Specifically, `AnalysisContext` will store type inference results for each method analyzed by `LSAnalyzer`,
> and `resolve_type` will be implemented as a query against these type inference results.
> However, this implementation requires JL to be able to lower arbitrary user code,
> which first requires integration of JL into Base.

Also adds `analyzer` field to `FullAnalysisResult` to support resolution queries.


aviatesk:
Now this PR allows signature help for cases like 
image

mlechu (approved):
Nice!

> `# TODO use JL once it supports general macro expansion`

Added to the mental list of places we need this

aviatesk:
> Added to the mental list of places we need this

Yeah, I think we'll need to integrate JL into Base sooner than expected.

PR #75, opened 2025-06-11
definition: deduplicate methods by function location

aviatesk:
Allows easier jump to essentially duplicated methods.

### DIFF
--- a/test/test_definition.jl
+++ b/test/test_definition.jl
@@ -111,6 +111,16 @@ include("setup.jl")
     #=34=# function say(h::Hel│lo)
     #=35=#     println("Hello, \$(hello.who)")
     #=36=# end
+    #=37=#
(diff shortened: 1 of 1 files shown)


src/definition.jl
```diff
@@ -90,7 +90,8 @@ function definition_target_methods(state::ServerState, uri::URI, pos::Position)
     objtyp = resolve_type(analyzer, mod, node)
     objtyp isa Core.Const || return empty_methods
 
-    return methods(objtyp.val)
+    # TODO modify this aggregation logic when we start to use more precise location informaiton
+    return unique(functionloc, methods(objtyp.val))
```
aviatesk, line 93:
```suggestion
    return unique!(functionloc, methods(objtyp.val))
```

PR #79, opened 2025-06-12
use `textEdit` in all global completion

abap34:
Changed all global completions to use `textEdit`. Previously, `insertText` caused inconsistent behavior across clients, leading to incorrect macro insertion. This change resolves the issue, as confirmed in multiple editors (VSCode, Helix).


abap34:
The difference in behavior between Helix and VSCode that I mentioned yesterday was caused by the `completionItem/resolve` processing. We were correctly sending `textEdit` in the `textDocument/completion` response, but not in the `completionItem/resolve` response. As a result, the client inserted an extra `@` due to its own word recognition logic.

I suspect that editors like Zed and VSCode respect the `textEdit` field from the `textDocument/completion` response, (I haven’t verified this, so take it with a grain of salt.) which is why the previous implementation appears to work.

aviatesk:
> We were correctly sending `textEdit` in the `textDocument/completion` response, but not in the `completionItem/resolve` response. As a result, the client inserted an extra `@` due to its own word recognition logic.

Makes sense. Good catch.

src/completions.jl
```diff
@@ -299,51 +299,50 @@ end
...
+        edit_start_pos = Position(; line=0, character=0)
+        is_macro_invoke = false
+    else
+        edit_start_pos = offset_to_xy(fi, fi.parsed_stream.tokens[current_token_idx - 1].next_byte)
+        is_macro_invoke = false
     end
+
+    create_edit(newText) = TextEdit(;
+        range = Range(; start = edit_start_pos, var"end" = pos),
+        newText = newText)
```
aviatesk: Let's avoid using capturing closures where possible.

src/completions.jl
```diff
@@ -299,51 +299,50 @@ end
...
+    current_token = fi.parsed_stream.tokens[current_token_idx]
+    current_kind = JS.kind(current_token)
+
+    if current_kind === JS.K"@"
+        edit_start_pos = offset_to_xy(fi, fi.parsed_stream.tokens[current_token_idx - 1].next_byte)
+        is_macro_invoke = true
+    elseif current_kind === JS.K"MacroName"
+        edit_start_pos = offset_to_xy(fi, fi.parsed_stream.tokens[current_token_idx - 2].next_byte)
+        is_macro_invoke = true
+    elseif current_kind === JS.K"TOMBSTONE"
```
aviatesk, line 323: What's `K"TOMBSTONE"` and why is it necessary?

src/utils.jl
```diff
@@ -275,3 +275,32 @@ function file_cache_error(uri::URI; data=nothing)
...
 end
+
+"""
+    get_current_token_idx(fi::FileInfo, offset::Int)
+
+Get the current token index at a given byte offset in a parsed file.
+This function returns the token at the specified byte offset, or `nothing`
+if the offset is invalid or no token exists at that position.
+
+In insertion mode, current means the token that locates right before the
```
aviatesk:
```suggestion
In insertion mode, "current" means the token that locates right before the
```

src/utils.jl
```diff
@@ -275,3 +275,32 @@ function file_cache_error(uri::URI; data=nothing)
...
+    findfirst(token -> token.next_byte >= offset, fi.parsed_stream.tokens)
+end
+
+get_current_token_idx(fi::FileInfo, pos::Position) =
+    get_current_token_idx(fi, xy_to_offset(fi, pos))
+
+
+get_current_token(fi::FileInfo, offset::Int) =
+    get_current_token_idx(fi, offset) === nothing ? nothing :
+    fi.parsed_stream.tokens[get_current_token_idx(fi, offset)]
```
aviatesk: This method seems to be no longer used anywhere.

src/completions.jl
abap34: `TOMBSTONE` is always at the beginning of the token stream.  This occurs when requesting completion for an empty program. I'll add a comment to clarify it.

src/utils.jl
abap34: fixed: 0bd269c

src/completions.jl
abap34: fixed: 0bd269c

src/completions.jl
abap34: 614679a

src/completions.jl
```diff
@@ -299,51 +299,52 @@ end
 
 function global_completions!(items::Dict{String, CompletionItem}, state::ServerState, uri::URI, params::CompletionParams)
     pos = params.position
-    is_macro_invoke = false
-    let context = params.context
-        if !isnothing(context)
-            # Don't trigger completion just by typing a numeric character:
-            context.triggerCharacter in NUMERIC_CHARACTERS && return nothing
```
aviatesk: Any reason for this removal?

abap34: I accidentally removed this along with the changs the macro recognition logic. 🙃
I'll add it back.

abap34: 2b2e215

src/completions.jl
```diff
@@ -437,7 +442,6 @@ function add_emoji_latex_completions!(items::Dict{String,CompletionItem}, state:
             kind=CompletionItemKind.Snippet,
             documentation=val,
             sortText,
-            filterText = key,
```
aviatesk: Is this no longer necessary?

abap34: It looks like I made a mistake while rebasing my previous commits. :pray:  I've fixed it.

aviatesk: While it's fine when code reviews and tests can catch these issues like in this case, some things are not easily testable, so that's not always guaranteed.
As a general practice, please review your code once more after submitting a PR:)

src/completions.jl
```diff
@@ -298,52 +298,58 @@ end
...
+                kind = CompletionItemKind.Variable,
+                documentation = nothing,
+                sortText = get_sort_text(0, #=isglobal=#true),
+                data = CompletionData(#=needs_resolve=#true),
+                labelDetails = CompletionItemLabelDetails(description = startswith(s, "@") ? "macro" : "global"),
+                textEdit = TextEdit(;
+                    range = Range(;
+                        start = edit_start_pos,
+                        var"end" = pos),
+                    newText = s))
```
```suggestion
            label = s,
            kind = CompletionItemKind.Variable,
            documentation = nothing,
            sortText = get_sort_text(0, #=isglobal=#true),
            data = CompletionData(#=needs_resolve=#true),
            labelDetails = CompletionItemLabelDetails(description = startswith(s, "@") ? "macro" : "global"),
            textEdit = TextEdit(;
                range = Range(;
                    start = edit_start_pos,
                    var"end" = pos),
                newText = s))
```

aviatesk:
LGTM. Thanks for working on this!

src/utils.jl
```diff
@@ -183,9 +183,9 @@ xy_to_offset(fi::FileInfo, pos::Position) = xy_to_offset(fi.parsed_stream.textbu
 
 Convert a 1-based byte offset to a 0-based line and character number
 """
-function offset_to_xy(ps::JS.ParseStream, byte::Int)
+function offset_to_xy(ps::JS.ParseStream, byte::Integer)
     # ps must be parsed already
-    @assert byte in JS.first_byte(ps):JS.last_byte(ps) + 1
+    @assert byte in JS.first_byte(ps):JS.last_byte(ps) + 1 "Byte offset $byte is out of bounds for the parse stream with first byte $(JS.first_byte(ps)) and last byte $(JS.last_byte(ps) + 1)"
```
aviatesk, line 188: Just for future reference: let's use `lazy"..."` for good reasons.

PR #81, opened 2025-06-12
defintion: follow up for aviatesk/JETLS.jl#77

aviatesk:
Properly support cases like:
```julia
let
    func│
end
```

### DIFF
--- a/src/definition.jl
+++ b/src/definition.jl
@@ -41,22 +41,24 @@ refs: https://github.com/rust-lang/rust-analyzer/blob/6acff6c1f8306a0a1d29be8fd1
 function select_target_node(st::JL.SyntaxTree, offset::Int)
     bas = byte_ancestors(st, offset)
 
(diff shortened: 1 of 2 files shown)


src/definition.jl
```diff
@@ -41,22 +41,24 @@ refs: https://github.com/rust-lang/rust-analyzer/blob/6acff6c1f8306a0a1d29be8fd1
...
-
     for i in 2:length(bas)
-        if kind(bas[i]) === K"."
-            # doesn't follow child module chain
-            if bas[i][1] === bas[i - 1]
-                return bas[i - 1]
-            end
+        basᵢ = bas[i]
+        if (kind(basᵢ) === K"." &&
+            basᵢ[1] !== target) # don't follow child module chain
```
aviatesk:
```suggestion
            basᵢ[1] !== target) # e.g. don't allow jumps to `tmeet` from `Base.Compi│ler.tmeet`
```

PR #86, opened 2025-06-13
completion: special case `@nospecialize`

aviatesk:
`@nospecialize` is very common in our code base and ignoring it completely
makes us unable to get completions for any methods using it, since it makes method signatures completely broken.

Fortunately JL can handle it (probably because JL special-cases it), so we can just keep it and support completions for such methods.

Note that this should be a temporary fix. Ideally we want our completions to handle any methods with any macros nicely, but it'd be possible after JL comes able to expand macros.

### DIFF
--- a/test/test_completions.jl
+++ b/test/test_completions.jl
@@ -310,7 +310,7 @@ end
         y
     end
     """
-    JETLS.cache_file_info!(state, uri, #=version=#1, text, filename)
+    JETLS.cache_file_info!(state, uri, #=version=#1, text)
     params = CompletionParams(;
         textDocument=TextDocumentIdentifier(; uri),
         position=Position(;line=1,character=6),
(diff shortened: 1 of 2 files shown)


test/test_completions.jl
```diff
@@ -389,7 +410,7 @@ end
             context=CompletionContext(;
                 triggerKind=CompletionTriggerKind.Invoked))
         items = JETLS.get_completion_items(state, uri, params)
-        @test any(items) do item
+        @test_broken any(items) do item
```
aviatesk:
```suggestion
        @test_broken any(items) do item # broken since JL currently seems unable to handle multi-arg `@nospecialize`
```
I found this test case is broken because JL fails to expand it and we're hitting this case:
https://github.com/aviatesk/JETLS.jl/blob/d15bc7ca959fae283c989683e7b0c40af6cd1b68/src/completions.jl#L150

mlechu: Might be worth opening an issue there

mlechu (approved):

test/test_completions.jl
aviatesk: It appears that JL is making custom definitions for some macros in Base to accommodate its macro expansion system without bootstrapping issue.
For now, we decided to adjust those custom definitions (https://github.com/c42f/JuliaLowering.jl/pull/15), but it seems we really need to improve the current situation regarding macro expansion with JL.

PR #88, opened 2025-06-13
completion: fix `greatest_lowerable` to return more minimal tree

aviatesk:
I had Claude implement the fix discussed at https://publish.obsidian.md/jetls/work/JETLS/Local+completion+performance+issue. It seems the performance issue was indeed caused by lowering on excessively large trees, and it appears to be fixed by tthis.

I'm not entirely confident that this fix is appropriate, so I would like to request a review.

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -71,46 +71,29 @@ end
 # =================
 
 """
-Find any largest lowerable tree containing the cursor and the cursor's position
-within it.  For local completions; something like least_unlowerable would be
-more helpful for globals.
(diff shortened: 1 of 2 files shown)


mlechu (commented):
I've pushed a more understandable update, though please check if it still fixes the performance.  We can also make it more strict if we want (e.g. just find the last function/let/loop in `bas`)

Thanks for letting me know that Claude wrote it, otherwise I would be a bit confused by the robot comments

aviatesk:
@mlechu Thank you very much for writing solid code. It really helps, and it does fix the performance issue.

PR #97, opened 2025-06-15
refactor source organization into modular architecture

aviatesk:
JETLS.jl and utils.jl file became huge, split the monolithic files into logical modules organized by functionality. Reorganizes code into analysis/, utils/, and feature-specific files while preserving all existing functionality. Also moves JSONRPC to its own directory as we do for LSP.jl.

/cc @mlechu @abap34 let me know if you have any objections.


aviatesk:
I don't want to keep this kind of large-scale refactor open for days, so I think I'll proceed as is for now once I confirm this passes test.
If you have any feedback, please feel free to let me know even after the merge.

PR #100, opened 2025-06-19
Distinguish `current_token_idx` and `prev_token_idx`

mlechu:
Bar cursors are a bit confusing; `abc|d` means our one-based byte offset is 4, but we mistakenly took that to mean `ab█d` instead of `abc█` in `current_token_offset`.  The examples were correct with respect to the description, but the function was actually returning the previous token idx.

Quick fix without tests; I know anything touching the token list needs to be refactored eventually due to JS changes anyway.

Also, let me know if any editors are inconsistent with this; I wouldn't be too surprised @aviatesk @abap34

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -325,11 +325,11 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
 
     # Case: `@│`
     if prev_kind === JS.K"@"
-        edit_start_pos = offset_to_xy(fi, fi.parsed_stream.tokens[prev_token_idx - 1].next_byte)
+        edit_start_pos = offset_to_xy(fi, JS.token_first_byte(fi.parsed_stream, prev_token_idx))
         is_macro_invoke = true
     # Case: `@macr│`
     elseif prev_kind === JS.K"MacroName"
-        edit_start_pos = offset_to_xy(fi, fi.parsed_stream.tokens[prev_token_idx - 2].next_byte)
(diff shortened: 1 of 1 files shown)


aviatesk:
Thanks, the changes look quite reasonable.
And I confirmed that it works without any problems in my editor as well.

> I know anything touching the token list needs to be refactored eventually due to JS changes anyway.

Do you have any idea what kind of changes will be needed?
For now we're pinning the JS version, but I was thinking we'll have to upgrade to the latest version eventually.

abap34 (approved):
It worked in my editor (Helix). I think the implementation is also fine.

mlechu:
> Do you have any idea what kind of changes will be needed?

ParseStream's `.tokens` is changed [here](https://github.com/JuliaLang/JuliaSyntax.jl/commit/36450b0d6c99e3b8ffa538ca74b1cd894901bf9c#diff-071c7153657f3dd094871f5c8eb4c1892473633dd07b47b0ef277143c3d1cfb2R437) (link doesn't work due to github not rendering this file... search for `tokens::` in parse_stream.jl), where the new `.output` is upgraded to encode the tree structure, but is no longer in source order. I think we'll just need to use a JuliaSyntax-provided iteration API instead of combing the list ourselves, so the functions changed in this PR would return a tree cursor (new structure) instead of an index.  In any case, it makes sense to keep it pinned in case more changes are needed to get the JS changes into Base, and I need to update JL before we can unpin it anyway.

aviatesk:
Sounds like a plan. Thanks for clarifying it.

PR #102, opened 2025-06-19
Add comment scope handling to global completions

aviatesk:
When completion is triggered within comments (, which seems to happen on some clients like Zed), set `textEdit` to `nothing` and let the client handle positioning based on `label` rather than attempting to calculate proper text edit ranges by ourselves.

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -335,6 +335,12 @@ function global_completions!(items::Dict{String, CompletionItem}, state::ServerS
     elseif current_kind === JS.K"TOMBSTONE"
         edit_start_pos = Position(; line=0, character=0)
         is_macro_invoke = false
+    elseif current_kind === JS.K"Comment"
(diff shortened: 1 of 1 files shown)


abap34 (approved):
LGTM

abap34:
It would be useful to be able to do LaTeX/emoji completions in comments and docstrings, but it looks  difficult without a some modification to JuliaSyntax :(

aviatesk:
> It would be useful to be able to do LaTeX/emoji completions in comments and docstrings, but it looks difficult without a some modification to JuliaSyntax :(

Handling such cases using JuliaSyntax would probably just add unnecessary complexity (it's not realistic to expect JS to parse comment text).
I think a very simple completion mode approach would be better, e.g., triggering LaTeX completion if there's a `\` between the current position and the preceding whitespace.

aviatesk:
xref: https://github.com/aviatesk/JETLS.jl/pull/103

PR #103, opened 2025-06-19
completion: support LaTeX/emoji completion in comment scope

aviatesk:
Enables backslash-triggered completions within comment scope by
searching backwards from cursor position for the nearest backslash
within the same comment token. Give up when there are preceding
whitespace before finding a backslash.

Also fixed the test cases to use `sizeof` to get the byte of input text
correctly.

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -381,8 +381,6 @@ end
 # LaTeX and emoji completions
 # ===========================
 
-# TODO allow LaTeX/emoji completions to be triggered within comment scope?
-
 """
     get_backslash_offset(state::ServerState, fi::FileInfo, pos::Position) -> offset::Int, is_emoji::Bool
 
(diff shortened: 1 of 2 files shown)


mlechu (approved):
Looks good. 

~~Just curious, is there anything stopping us from using the same character-based approach in the general case? I'm not sure if tokenizing is even necessary~~ Actually, I realize that most people have completions show up automatically, and I think they would show up for backslashes in strings, which would be annoying

aviatesk:
Yeah, I think the token-based approach is good when possible, because it allows completions to be triggered considering language semantics to some extent.
But, especially in cases like this one, I also think we don't need to stick to it too strictly.
Basically, I think it's fine to trigger completion based on token analysis where possible and add a naive but simple implementation like this when that can't be applied easily.

Even though it increases implementation complexity.

PR #104, opened 2025-06-19
definition: handle methods with unknown locations

abap34:
Currently, Go to Definition fails for `+` because of:

```julia
julia> m = methods(+, (LinearAlgebra.UniformScaling, LinearAlgebra.Diagonal))[1]
+(B::UniformScaling, A::Diagonal)
     @ LinearAlgebra none:0

julia> functionloc(m)
ERROR: could not determine location of method definition
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:44
 [2] functionloc(m::Method)
   @ Base ./methodshow.jl:173
 [3] top-level scope
   @ REPL[30]:1
```

(https://github.com/JuliaLang/LinearAlgebra.jl/blob/c5d712277675649d79fd8921eec0b78587e7f72b/src/special.jl#L271-L273)

This change skips such methods instead of throwing them as errors. This lookup strategy will be removed in the future, but it will remain for a while, so I made a simple fix.

(I'm now using `+` as an example of a case where it fails,  but let me know if there’s a more appropriate one that is likely to remain with a unknown location for a longer time)

### DIFF
--- a/src/definition.jl
+++ b/src/definition.jl
@@ -61,6 +61,8 @@ function select_target_node(st::JL.SyntaxTree, offset::Int)
     return nothing
 end
 
+is_location_unknown(m) = Base.updated_methodloc(m)[2] <= 0
+
 """
 Get the range of a method. (will be deprecated in the future)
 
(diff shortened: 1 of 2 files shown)


aviatesk:
LGTM.

> (I'm now using `+` as an example of a case where it fails, but let me know if there’s a more appropriate one that is likely to remain with a unknown location for a longer time)

`+` seems to be a good candidate.

aviatesk (approved):

abap34:
(fixed typo)

aviatesk:
Good. Feel feel to merge this yourself when you think it's ready.

PR #105, opened 2025-06-19
completion: simplify backslash completion implementation

aviatesk:
After reconsidering the discussion at https://github.com/aviatesk/JETLS.jl/pull/103#pullrequestreview-2943628486, now I think it's better to have this completion even inside string scopes. If the completion shows up, you can just ignore it, so it'd be better than not having it available.

My previous comment was based on the fact that implementing general completion features like `global_completions!` requires tokenization, but for this specific backslash completion, I believe tokenization is not really needed.


mlechu (approved):
Sounds reasonable, and it's a nice simplification.  I'll turn on the automatic completion popup so I have a better idea of what people see.

PR #106, opened 2025-06-19
Signature help: Macro support and other improvements

mlechu:
- Fix `call1() = call2(|)`, where signature help didn't work before since it
  triggered our function declaration detection
- Fix declaration detection with `where` (silly iteration bug; I wasn't testing
  this properly)
- Explicitly ignore infix and postfix calls.  Before, we were assuming the first
  child of the call is always the function, but this is only true for prefix
  calls, so we failed to show help by chance.  I chose to ignore these mostly to
  avoid adding more edge case code, but I also think it's an OK choice
  taste-wise---we'll only choose to show help for `f` everywhere in
  ```julia
  f(x + y, a:b, M')
  ```
  Can always revisit.
- Basic support for dotcall (just treat it like a normal call)
- Make signature help work with macro calls
  - `@m(a,b,c)` and `@m a b c` supported
  - Hide the first two `__module__` and `__source__` parameters with a regex hack
- (most of the code in this PR, but necessary for unparenthesized macro calls)
  Detect whitespace behind the cursor and show signature help on an "if we typed
  something here" basis. i.e. support for the following cases:
  ```julia
  foo(#=odd_comment, no closing paren=#█
  ```
  
  ```julia
  foo(long_________________arg,
      x█
  ```

  ```julia
  @m(
     #=same idea=# █
  ```

  ```julia
  @m #=but don't show help if there's a newline after this=# █
  ```
- pile of tests

### DIFF
--- a/src/signature-help.jl
+++ b/src/signature-help.jl
@@ -157,7 +157,7 @@ function compatible_method(m::Method, ca::CallArgs)
         msig = sprint(show, m; context=(:compact=>true, :print_method_signature_only=>true))
     else
         mstr = sprint(show, m; context=(:compact=>true))
-        msig_locinfo = split(mstr, '@')
+        msig_locinfo = split(mstr, " @ ")
         length(msig_locinfo) == 2 || return false
         msig = strip(msig_locinfo[1])
     end
@@ -228,7 +228,7 @@ function make_siginfo(m::Method, ca::CallArgs, active_arg::Union{Int, Symbol};
         msig = sprint(show, m; context=(:compact=>true, :print_method_signature_only=>true))
     else
         mstr = sprint(show, m; context=(:compact=>true))
-        msig_locinfo = split(mstr, '@')
+        msig_locinfo = split(mstr, " @ ")
         length(msig_locinfo) == 2 || return false
         msig = strip(msig_locinfo[1])
     end
--- a/src/signature-help.jl
+++ b/src/signature-help.jl
@@ -10,6 +10,7 @@ signature_help_options() = SignatureHelpOptions(;
 
 const SIGNATURE_HELP_REGISTRATION_ID = "jetls-signature-help"
 const SIGNATURE_HELP_REGISTRATION_METHOD = "textDocument/signatureHelp"
+const CALL_KINDS = KSet"call macrocall dotcall"
 
 function signature_help_registration()
     (; triggerCharacters, retriggerCharacters) = signature_help_options()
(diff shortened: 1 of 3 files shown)


mlechu (commented):
`resolve_type` in `cursor_siginfos` appears to be having trouble with macro calls, but only on 1.12

aviatesk:
> `resolve_type` in `cursor_siginfos` appears to be having trouble with macro calls, but only on 1.12

Fixed at c4ffa4f. It turned out that it's because the `sprint(show, m)` hack needed more tweak on v1.12.

aviatesk (approved):
Thank you for implementing this wonderful enhancement.
Overall, I think this is very good code. I've read through the code and it seems like all the edge cases I can think of seem to be handled properly.

> I chose to ignore these mostly to avoid adding more edge case code

I agree. I don't think showing signature help for infix calls is very useful in most cases. While it's convenient that signature help allows us to understand matching methods, in many cases the actually useful information there is to find cases where there are no matching methods, and reporting such cases is the role of diagnostics. In other words, I think signature help for infix calls could rather become cumbersome and worsen the UX. Of course, from an algorithmic complexity perspective, this approach is undoubtedly better.

If I were to comment on one point regarding macro signature help, it might be good to add `" "` to `triggerCharacters`. It would be better to get signature help when you type `@macrofunc |` (currently we need to invoke it manually). By registering `" "`, signature help requests will be called quite frequently, but I think this algorithm is written to be quite fast, so it should be fine.

src/signature-help.jl
```diff
@@ -154,10 +157,14 @@ function compatible_call(m::Method, ca::CallArgs)
...
     else
         mstr = sprint(show, m; context=(:compact=>true))
-        msig_locinfo = split(mstr, '@')
+        msig_locinfo = split(mstr, " @ ")
         length(msig_locinfo) == 2 || return false
         msig = strip(msig_locinfo[1])
     end
+    if ca.kind === K"macrocall" # hack. TODO delete
+        msig = replace(msig, "__source__::LineNumberNode, __module__::Module, "=>"",
+                       "__source__::LineNumberNode, __module__::Module"=>""; count=1)
```
aviatesk, line 166: Any reason for why we can't simply `replace(msig, "__source__::LineNumberNode, __module__::Module, "=>""; count=1)`?

mlechu: Second case is for zero-arg macros.  Kind of gross, I know, but this line will go away when we improve `compatible_method`, and copy of this line we use for creating the label isn't worse than what we would have to do otherwise.

aviatesk: Oh I just didn't notice the `, ` part of the first case. Thanks for your explanation.

PR #109, opened 2025-06-22
signature-help: some minor tweaks

aviatesk:


aviatesk:
> Agree we can disable help for large code "arguments" to macro calls, but not sure `block` is the only thing to look for in the heuristic. Would it work to hide signature help if the cursor isn't on the same line as the `@` (maybe only in the unparenthesized case)?

Yeah, I agree that whether it's a `block` doesn't seem like a generalized and reliable condition.
In the unparenthesized macrocall case, requiring that the cursor exists on the same line as the `@` sounds to be a simpler and better condition.

PR #111, opened 2025-06-23
implement "hover" for global symbols

aviatesk:
Just to show the documentation of global symbols.

https://github.com/user-attachments/assets/eb1f76ca-0a4d-46b0-8151-addd601417b2


abap34:
This is obviously not the issue that this PR is supposed to fix (and it's a minor issue with low urgency, not something we need to tackle right now 🙂), but I wanted to leave a note about something I noticed.

In some editors, the syntax highlighting for `jldoctest` code blocks doesn't render well, which negatively affects the appearance. 

image


Applying a small hack such as rewriting `blk isa Markdown.Code && blk.language === "jldoctest"` to change the `language` field to `"julia"` can slightly improve the appearance, though of course it's not a complete solution.

image


Rendering proper highlighting should be the responsibility of the client, and ultimately depends on the parser used by the client. It seems that a dedicated parser for `julia-repl` in editor may be necessary for proper support. (I looked around but couldn't find one.) Parsers for regular Julia code already exist, so I don't think it would be too hard to create one that produces the same results as `Makdown.term`.

aviatesk:
> Applying a small hack such as rewriting `blk isa Markdown.Code && blk.language === "jldoctest"` to change the `language` field to `"julia"` can slightly improve the appearance, though of course it's not a complete solution.

I agree, it's difficult to expect all clients to have such specialized syntax highlighters, so I think it would be better to adjust the highlight language on the server side.
I think it's a good idea to prepare a traverser for `Markdown.MD` objects and rewrite the `language` of `CodeBlock`.
Related to this, we will also need to change `[...](@ref)` to links that clients can actually handle in the future.

aviatesk:
There are many things remain to be resolved in the future, but I'm going to merge this for now.

aviatesk:
The remaining issues are tracked at https://publish.obsidian.md/jetls/work/JETLS/Hover.

aviatesk:
The markdown issues are tracked at https://publish.obsidian.md/jetls/work/JETLS/Markdown+issues.

PR #115, opened 2025-06-24
Implement Go to Definition for local bindings

abap34:
https://github.com/user-attachments/assets/cb924339-4475-4171-b390-db3c88ddc257

This is a prototype implementation of jumping to local bindings.
It is based on the bindings information provided by JuliaLowering, similar to the current local completion mechanism.

## Known limitations

As with completion, this implementation suffers from the following limitations:

1. Only the first assignment (i.e., where the binding is created) is shown, even when a variable is reassigned later.

```julia
function redef(x)
    x = 1
    y = x│ + 1      # Ideally, both the argument and `x = 1` should be shown
    return y
end
```

2. It does not work correctly when the execution order does not align with the position in the code.

Currently, bindings are filtered simply based on cursor position. 

https://github.com/aviatesk/JETLS.jl/blob/415963d3c07b6454e8f83b93625cae5b0580dd3e/src/completions.jl#L99-L115

As a result, in the following case, `x` is not found:

```julia
function not_linear()
    finish = false
    @label l1
    !finish && @goto l2
    return x│
    @label l2
    x = 1
    finish = true
    @goto l1
end
```

In addition, there are other cases where it fails as well; these are currently marked as broken in test_definition.jl.

I think this PR can be reviewd && merged after rebase some of the preceding PRs related to definitions.

Also, the test cases have become bloated and now require heroic effort to add new cases 🙃.
 After this PR is merged, I plan to rework them into a more human-maintainable.


aviatesk:
This seems to be great work:)

> I think this PR can be reviewd && merged after rebase some of the preceding PRs related to definitions.
> 
> Also, the test cases have become bloated and now require heroic effort to add new cases 🙃. After this PR is merged, I plan to rework them into a more human-maintainable.

Sounds like a plan. I'd give a review on this once the rebase is done. Also looking forward to the test refactor (maybe in another PR as you suggested).

src/definition.jl
```diff
@@ -51,6 +59,50 @@ function module_definition_location(mod::Module)
             var"end" = Position(; line = line - 1, character = Int(typemax(Int32)))))
 end
 
+get_location(bind::Tuple{JL.SyntaxTree, URI}) =  Location(; uri = bind[2], range = get_source_range(bind[1]))
+
+function definition_target_methods(objtyp)
```
aviatesk: As discussed earlier, this PR will be refactored so that the caller side handles case-by-case branching for each type of `objtyp`, but I'll comment on this for future reference.

In implementations like this one that dispatch based on lattice objects returned by the compiler, the compiler generates dynamic dispatch because the type of `objtyp` cannot be statically determined from the caller side either.
In such cases, using `@nospecialize` can enable static dispatch.

abap34:
Refactoring and following upstream changes.
(This also includes a minor spec change we discussed.)

**Local bindings are now prioritized** in search results.
   As a result, in cases like the following, `Base.sin` no longer appears as a candidate, only the local `sin` appear:

```julia
function f(x)
    sin = x -> x * x
    sin(2)
end
```


There are also some minor fixes that affect completions.
Currently, when `│func(1.0)` appears at the toplevel, `greatest_local` treats  `(call func 1.0)` as the largest lowerable tree. As a result, a `func` binding is created at the call site, which is prioritized during Go to Definition (because it's considered a local binding).

Ideally, `greatest_local` should return `nothing` for such a top-level call, but for now I solve this problem by having `is_relevant` return `false` when the cursor is in an identifier.
This range did not cover the left side of the identifier, which was a problem in this PR, so I have fixed it. This will allow go-to-definition to work correctly in such edge cases.

This also changes the behavior of completions slightly: triggering completion at `|func(1.0)` for an undefined function `func` will no longer show `func` as a suggestion.

src/utils/binding.jl
```diff
@@ -0,0 +1,102 @@
...
+    !binding.is_internal &&
+        !in(cursor, start:(stop+1)) &&
+        (binding.kind === :global
+         # || we could relax this for locals defined before the end of the
+         #    largest for/while containing the cursor
+         || cursor > start)
+end
+
+let lowering_module = Module()
+    global function jl_lower_for_completion(st0)
```
aviatesk:
```suggestion
    global function jl_lower_for_scope_resolution(st0)
```
Maybe we should rename this function now.

src/definition.jl
```diff
@@ -71,49 +105,47 @@ function handle_DefinitionRequest(server::Server, msg::DefinitionRequest)
...
-    if !(objtyp isa Core.Const)
-        return send(server, DefinitionResponse(; id = msg.id, result = null))
-    end
 
-    obj = objtyp.val
-    if obj isa Module
-        modloc = module_definition_location(obj)
-        if modloc === nothing
+    objtyp === nothing && return send(server, DefinitionResponse(; id = msg.id, result = null))
+    objtyp isa Core.Const || return send(server, DefinitionResponse(; id = msg.id, result = null))
```
aviatesk:
```suggestion
    objtyp isa Core.Const || return send(server, DefinitionResponse(; id = msg.id, result = null))
```

src/definition.jl
```diff
@@ -71,49 +105,47 @@ function handle_DefinitionRequest(server::Server, msg::DefinitionRequest)
...
+
+    if objtyp.val isa Module
+        if is_location_unknown(objtyp.val)
```
aviatesk:
```suggestion
    objval = objtyp.val
    if objval isa Module
        if is_location_unknown(objval)
```
Use this pattern for maximum type stability and invalidation robustness.
`LSAnalzyer` can analyze this kind of code without any problem (since it enables the `MustAlias` lattice element), but `NativeInterpreter` doesn't enable the ability to infer it.

src/utils/binding.jl
```diff
@@ -0,0 +1,102 @@
...
+        # find the innermost hard scope containing this binding decl.  we shouldn't
+        # be in multiple overlapping scopes that are not direct ancestors; that
+        # should indicate a provenance failure
+        i = findfirst(ba -> JS.kind(ba) in KSet"scope_block lambda module toplevel", bas)
+        push!(bscopeinfos, (binfo, bas, isnothing(i) ? nothing : bas[i]))
+    end
+
+    cursor_scopes = byte_ancestors(st2, b)
+
+
```
aviatesk:
```suggestion
```

src/utils/binding.jl
```diff
@@ -0,0 +1,102 @@
...
+"""
+function cursor_bindings(st0_top::JL.SyntaxTree, b_top::Int)
+    st0, b = greatest_local(st0_top, b_top)
+    if isnothing(st0)
+        return nothing # nothing we can lower
+    end
+    ctx3, st2 = try
+        jl_lower_for_completion(st0)
+    catch err
+        JETLS_DEV_MODE && @warn "Error in lowering" err
```
aviatesk:
```suggestion
        # JETLS_DEV_MODE && @warn "Error in lowering" err
```

src/utils/binding.jl
abap34: I agree. resolved  f16724a

src/definition.jl
```diff
@@ -51,6 +53,38 @@ function module_definition_location(mod::Module)
...
+    # because we only look for local bindings
+    kind(node) !== K"Identifier" && return nothing
+    cbs = cursor_bindings(st, offset)
+    (cbs === nothing || isempty(cbs)) && return nothing
+    matched_binding = findfirst(b -> b[1].name == node.name_val, cbs)
+    matched_binding === nothing && return nothing
+
+    # Will return multiple results when we support tracking of multiple assignments.
+    # `SyntaxTree` does not retain URI information because `ParseStream` does not store it
+    # so we explicitly keep it.
```
aviatesk, line 69: We can construct `URI` via `filename2uri(JS.filename(JS.sourcefile(cbs[matched_binding][2])))`, but I agree with that just propagating `uri` object would be simpler. I don' think we need to pass `uri` object to this function from the first place though.

aviatesk:
> There are also some minor fixes that affect completions. Currently, when `│func(1.0)` appears at the toplevel, `greatest_local` treats `(call func 1.0)` as the largest lowerable tree. As a result, a `func` binding is created at the call site, which is prioritized during Go to Definition (because it's considered a local binding).
> 
> Ideally, `greatest_local` should return `nothing` for such a top-level call, but for now I solve this problem by having `is_relevant` return `false` when the cursor is in an identifier. This range did not cover the left side of the identifier, which was a problem in this PR, so I have fixed it. This will allow go-to-definition to work correctly in such edge cases.
> 
> This also changes the behavior of completions slightly: triggering completion at `|func(1.0)` for an undefined function `func` will no longer show `func` as a suggestion.

I think this change is good, but it might be better to merge it in a separate PR first, along with a test case for `|func(1.0)` where `func` isn't defined.

src/definition.jl
abap34: I think `filename2uri(JS.filename(JS.sourcefile(cbs[matched_binding][2]))` doesn't work (`JS.filename` return empty string) because `parse_stream` is simply made from text and no file information is passed.

https://github.com/aviatesk/JETLS.jl/blob/77c74d011df003233c4fc76971f14c76cd1244b7/src/document-synchronization.jl#L1-L5

So I'm making it explicitly retain it.

src/definition.jl
```diff
@@ -71,49 +105,47 @@ function handle_DefinitionRequest(server::Server, msg::DefinitionRequest)
...
-                targetRange = modloc.range,
-                targetSelectionRange = modloc.range,
-                originSelectionRange)]
-            return send(server, DefinitionResponse(; id = msg.id, result))
         else
-            return send(server, DefinitionResponse(; id = msg.id, result = modloc))
+            return send(server,
+                DefinitionResponse(;
+                    id = msg.id,
+                    result = create_definition([objtyp.val],
```
aviatesk:
```suggestion
                    result = create_definition([objtyp.val],
```
```suggestion
                    result = create_definition([objval],
```

src/definition.jl
```diff
@@ -51,6 +53,38 @@ function module_definition_location(mod::Module)
...
+            LocationLink(;
+                targetUri = loc.uri,
+                targetRange = loc.range,
+                targetSelectionRange = loc.range,
+                originSelectionRange = origin_range)
+        end
+    else
+        get_location.(objects)
+    end
+end
```
aviatesk:
```suggestion
function create_definition(object, origin_range::Range, locationlink_support::Bool)
    loc = @inline get_location(obj)
    if locationlink_support
        LocationLink(;
            targetUri = loc.uri,
            targetRange = loc.range,
            targetSelectionRange = loc.range,
            originSelectionRange = origin_range)
    else
        loc
    end
end
```
and broadcasting from callsite would be cleaner, allowing to cut off the allocation of `Vector{Module}` as a micro optimization?

src/definition.jl
abap34: resolved f039e3f

src/definition.jl
abap34: I agree. resolved f039e3f

abap34:
Thank you for the review! I’ll split out the parts that affect completion into a separate PR and open it.

(#123)

aviatesk (approved):
LGTM. Thanks so much for your efforts! :heart:

PR #123, opened 2025-06-26
exclude identifier from bindings when cursor is at its left edge

abap34:
With this change, when `func` is undefined, it will no longer appear as a completion candidate in cases like:

```julia
│func(1.0)
```

Similarly, in Go to Definition, the `func` on the right side of the cursor will no longer be considered as a definition candidate.

ref: https://github.com/aviatesk/JETLS.jl/pull/115#issuecomment-3007784009


mlechu (approved):
Looks good, thanks for fixing

PR #124, opened 2025-06-26
signature-help: disable sig help within `do` block

aviatesk:


mlechu:
Thanks, I didn't realize `do` blocks also included a call expression

Assuming it always parses to `(call f (do (tuple ...) ...))`, would it be easier to add this check to `is_relevant_call`? That would also make the `do` call transparent in a case like this
```
f1(identity() do
   |
   end) # shows help for `f1`
```

aviatesk:
I tried something like:
```diff
diff --git a/src/signature-help.jl b/src/signature-help.jl
index ae61ab5..0545992 100644
--- a/src/signature-help.jl
+++ b/src/signature-help.jl
@@ -342,6 +342,17 @@ function is_crossline_noparen_macrocall(call::JL.SyntaxTree, cursor_byte::Int)
     end
 end
 
+function is_relevant_call(bas::JL.SyntaxList, i::Int)
+    st = bas[i]
+    is_relevant_call(st) || return false
+    if i ≥ 2
+        if JS.kind(bas[i-1]) == JS.K"do"
+            return false
+        end
+    end
+    return true
+end
+
 """
 Return the nearest call in `st0` containing cursor byte b (if any).
 
@@ -358,7 +369,7 @@ function cursor_call(ps::JS.ParseStream, st0::JL.SyntaxTree, b::Int)
     end
 
     bas = byte_ancestors(st0, b)
-    i = findfirst(is_relevant_call, bas)
+    i = findfirst(i::Int->is_relevant_call(bas, i), 1:length(bas))
     if !isnothing(i)
         basᵢ = bas[i]
         if call_is_decl(bas, i, basᵢ)
@@ -369,9 +380,6 @@ function cursor_call(ps::JS.ParseStream, st0::JL.SyntaxTree, b::Int)
             #     ... | ...
             # end
             return nothing
-        elseif any(j::Int->JS.kind(bas[j])===JS.K"do", 1:i)
-            # bail out if this is actually within a `do` block
-            return nothing
         end
         return basᵢ
     end
```

However, I've come to the conclusion that having signature help appear in such situations is indeed bothersome. Since we currently have `" "` registered in `triggerCharacters`, signature help occurs quite frequently within `do` blocks, so I think it would be fine to disable signature help for those caller call(s) within `do` blocks.

Screenshot 2025-06-27 at 05 11 40

mlechu:
Always disabling help in the do block sounds reasonable.  If it comes up again, we might want to adjust what we do when triggered by " ".

aviatesk:
Thanks for your review!

PR #125, opened 2025-06-26
definition: enhance local binding jump

aviatesk:
This change enhances the local binding jump functionality to support multiple assignment operations and enables jumps from target bindings that are positioned before the definition.

@abap34 Sorry for immediately fixing the code that was reviewed and merged. There were reasons for making the changes, so I'll explain it in inline comments. I would appreciate it if you could review it.

### DIFF
--- a/src/definition.jl
+++ b/src/definition.jl
@@ -33,7 +33,7 @@ Get the range of a method. (will be deprecated in the future)
 TODO (later): get the correct range of the method definition.
 For now, it just returns the first line of the method
 """
-function get_location(m::Method)
+function LSP.Location(m::Method)
     file, line = functionloc(m)
     file = to_full_path(file)
     return Location(;
@@ -43,7 +43,7 @@ function get_location(m::Method)
             var"end" = Position(; line = line - 1, character = Int(typemax(Int32)))))
 end
 
-function get_location(mod::Module)
+function LSP.Location(mod::Module)
     (; file, line) = Base.moduleloc(mod)
     uri = filename2uri(to_full_path(file))
     return Location(;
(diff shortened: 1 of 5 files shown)


src/definition.jl
```diff
@@ -98,52 +94,62 @@ function handle_DefinitionRequest(server::Server, msg::DefinitionRequest)
...
+            result = LocationLink[
+                LocationLink(loc, get_source_range(target_binding))
+                for loc in result]
+        end
         return send(server,
             DefinitionResponse(;
                 id = msg.id,
-                result = create_definition.(lbs,
-                    Ref(originSelectionRange),
-                    Ref(locationlink_support))))
```
aviatesk, line 116: While the previous implementation was simpler and cleaner, since broadcasting results depend on type inference, I encountered the following constructor error during the refactoring process:
```
ERROR: MethodError: Cannot `convert` an object of type
  Vector{Any} to an object of type
  Union{JETLS.LSP.Null, Vector{JETLS.LSP.Location}, Vector{JETLS.LSP.LocationLink}, JETLS.LSP.Location}
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base_compiler.jl:133

Stacktrace:
  [1] DefinitionResponse
    @ ~/julia/packages/JETLS/src/LSP/language-features/definition.jl:29 [inlined]
```
While not as elegant, I have changed to more robust code using typed vcat and easy-to-infer constructs.

src/definition.jl
```diff
@@ -53,36 +53,32 @@ function get_location(mod::Module)
...
-    cbs = cursor_bindings(st, offset)
-    (cbs === nothing || isempty(cbs)) && return nothing
-    matched_binding = findfirst(b -> b[1].name == node.name_val, cbs)
-    matched_binding === nothing && return nothing
-
-    # Will return multiple results when we support tracking of multiple assignments.
-    # `SyntaxTree` does not retain URI information because `ParseStream` does not store it
-    # so we explicitly keep it.
-    return [(cbs[matched_binding][2], uri)]
-end
```
aviatesk, line 71: To find multiple definitions, it turned out that we need to traverse `st3` rather than using `cursor_bindings`, so that's what I'm doing.

test/test_definition.jl
```diff
@@ -18,15 +19,40 @@ myidentity(x) = x
...
+                for x in xs]
+        end
+    """) do res
+        @test !isnothing(res)
+        binding, defs = res
+        @test JS.source_line(JL.sourceref(binding)) == 2
+        @test length(defs) == 1
+        @test JS.source_line(JL.sourceref(only(defs))) == 3
+    end
+end
```
aviatesk: I'd like to leave this refactor to you, but I think we may want to refactor the `"'Definition' for local bindings request/responce"` test set into more minimal test cases by using subroutines like `with_local_definitions`.

aviatesk: NVM, Claude does the refactoring for us f1e7941.

src/definition.jl
```diff
@@ -98,52 +94,62 @@ function handle_DefinitionRequest(server::Server, msg::DefinitionRequest)
...
 
     objval = objtyp.val
+    originSelectionRange = get_source_range(node)
     if objval isa Module
         if is_location_unknown(objval)
             return send(server, DefinitionResponse(; id = msg.id, result = null))
         else
+            local result = Location(objval)
+            if locationlink_support
+                result = LocationLink(result, originSelectionRange)
```
abap34: Surprisingly, it seems that they do not accept a single `LocationLink` as a valid response (even though a single `Location` is permitted 😇) 

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition:~:text=DefinitionParams%20extends%20TextDocumentPositionParams%2C%0A%09WorkDoneProgressParams%2C%20PartialResultParams%20%7B%0A%7D-,Response%3A,-result%3A%20Location%20%7C%20Location%5B%5D%20%7C%20LocationLink%5B%5D%20%7C%20null

aviatesk: Good catch!

PR #129, opened 2025-06-27
Add JL utils for interactive debugging

mlechu:
I noticed some convenience functions making it into `/test`, so here are some other ones I have lying around.  If it doesn't make sense to have these here, feel free to close this PR and take them for your own startup.jl :)


aviatesk (approved):
`jldebug` looks useful.
Thanks for sharing the code!

PR #134, opened 2025-07-01
test: refactor test_global_definition to use do-block syntax

aviatesk:
Refactor the test_global_definition helper function to take the tester function as the first argument, enabling the use of Julia's cleaner do-block syntax. This change follows the more canonical Julia coding style and improves readability of the test cases.

🤖 Generated with [Claude Code](https://claude.ai/code)


aviatesk:
@abap34 Just pinged you for reference, wondering if you agree with this refactoring.

PR #143, opened 2025-07-03
test: refactor completion and signature-help test utilities

abap34:
I updated the tests for completion and signature help to follow the same approach as the other tests. Some custom utility functions have also been replaced with the shared ones.

### DIFF
--- a/test/test_completions.jl
+++ b/test/test_completions.jl
@@ -104,17 +104,22 @@ module M
     end
 end
 """
+    cnt = 0
     with_completion(code) do i, cv
         if i == 1
             cv_nhas(cv, ["a", "b", "c", "x", "y", "z"])
+            cnt += 1
         elseif i == 2
             cv_has(cv, ["x", "y", "z"], kind=:local)
             cv_nhas(cv, ["a", "b", "c"])
+            cnt += 1
         elseif i == 3
             cv_has(cv, ["a", "b", "c"], kind=:local)
             cv_nhas(cv, ["x", "y", "z"])
+            cnt += 1
         end
     end
+    @test cnt == 3
 end
 
 @testset "nested and adjacent scopes" begin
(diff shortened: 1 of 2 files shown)


test/test_completions.jl
```diff
@@ -31,221 +38,232 @@ function cv_has(cs::Vector{CompletionItem}, expected, kind=nothing)
...
             local x
             x
         end
     end
 end
 """
-    test_cv(code, "#1", "x", kind=:global)
-    # broken. JuliaLowering bug?
-    # test_cv(code, "#2", "x", kind=:local)
+    with_completion(code) do i, cv
```
aviatesk, line 192: Use
```julia
cnt = 0
with_completion(code) do i, cv
    ... # increment `cnt` in each `if` clause
end
@test cnt == 2
```
pattern to ensure all the expected test cases are visited.

test/test_completions.jl
```diff
@@ -31,221 +38,232 @@ function cv_has(cs::Vector{CompletionItem}, expected, kind=nothing)
...
+        if i == 1
+            cv_has(cv, ["x"], kind=:global)
+        elseif i == 2
+            # broken. JuliaLowering bug?
+            # cv_has(cv, ["x"], kind=:local)
```
aviatesk, line 198: This is expected, as `cursor_bindings` try to exclude bindings that are defined after the cursor position.

aviatesk:
@abap34 Is it expected that this PR adds around 1min overhead to the whole test suite execution with 121 new cases added?

abap34:
Some of the tests are taking longer in CI because they've changed to full-cycle tests. I should have made that clearer. 🙏

I'll implement optimization by reusing the server instead of starting it up every time and some bug fixes soon. (also look into the number of tests being run.)

abap34:
It looks like the increase in the number of tests is due to the same reason. (Tests in `withserver`)
https://github.com/aviatesk/JETLS.jl/blob/5c7756da19c546365d0061bae151dc12785d8e4d/test/setup.jl#L145-L150

By the way, I’ve found a few bugs and optimization point, so I’ll fix those.

aviatesk:
Ok, so it's expected. Just wanted to confirm. It would be awesome if we can optimize the test execution.

abap34:
What do you think about running the tests in parallel?

I believe the assumption that the `test_{features}.jl` files don't interfere with each other will likely remain valid in the future as well.

aviatesk:
Yeah, I agree with the parallelization.

PR #145, opened 2025-07-03
hover: improve Markdown appearance

abap34:
This PR introduces a custom Markdown transformation mechanism.

Currently, it only provides a rule to normalize code block languages, converting `julia-repl` and `jldoctest` to `julia`.
All other content falls back to the default output produced by `Markdown.plain`.

Ideally, we would also like to support constructs like `@ref` in the future. However, since this turned out to be more complex 
than expected, this PR deliberately leaves such cases unmodified for now.

Transformation rules can be easily extended by enhancing the `lsrender` in future.


abap34:
It looks like this branch is a bit behind master, so I'll rebase it.
(I'll request a review once it's done.)

pfitzseb:
> Ideally, we would also like to support constructs like `@ref` in the future. However, since this turned out to be more complex
than expected, this PR deliberately leaves such cases unmodified for now.

I'd suggest looking for existing bindings locally to cover some cases; everything else essentially needs the full Documenter context to resolve (e.g. links to a section in the docs), so there needs to be some mechanism to ship this kind of metadata (maybe based on https://github.com/JuliaDocs/DocumenterInterLinks.jl?tab=readme-ov-file#inventories).

abap34:
Thank you for the advice and the helpful link!

The biggest problem we’ve faced is that, within the scope of the LSP, it’s difficult to customize the behavior when a link in a Markdown document is clicked. We’d like to click a link to trigger some custom command like a Go to Definition, but since that would require some implementation on the client side, the issue has been deferred for now.

pfitzseb:
Well, you *could* always rewrite this to an external link to the documentation for whatever object is being referenced. The VS Code extension instead triggers code search, but that's obviously very specific to that client.

src/utils/server.jl
```diff
@@ -66,13 +67,49 @@ function get_context_module(analysis_unit::AnalysisUnit, uri::URI, pos::Position
...
+lsrender(io::IO, md::Markdown.MarkdownElement) = Markdown.plain(io, md)
+
+function lsrender(io::IO, md::Markdown.MD)
+    isempty(md.content) && return
+    for md in md.content[1:end-1]
+        lsrender(io, md)
+        println(io)
+    end
+    lsrender(io, md.content[end])
+end
```
aviatesk: I was thinking more along the lines of using something like `MacroTools.postwalk` to modify the `Markdown.MD` object itself,
and letting the print part be handled by the `show` method that `Markdown` implements. Have you considered that approach?

While the implementation of `lsrender` is quite simple so it's probably not much of an issue, I'd like to think about whether we should have our own `print` implementation.

aviatesk:
I want to make this LS as client-agnostic as possible, so the approach I'll probably take regarding `@ref` is to replace it with `[...](file://...)` or `[...](url://...)`.
This isn't fully compliant with LSP, but almost all language clients support such URI-jumping.
I would lean toward the former internal and local jump. However, in this case, we can't narrow down the target in cases like multiple methods, so it might be good to do that when the destination is uniquely determined, and otherwise use something like `window/showMessage`.

src/utils/server.jl
abap34: In fact, `plain` is the implementation of `show` for `MIME"text/markdown"`.

What currently looks like a manual implementation is actually almost identical to the one in `stdlib/Markdown`, so I don't think it will incur much maintenance cost.

I also tried implementing a traversal approach, but since each `MarkdownElement` has quite a different structure, I think the current implementation to be simpler than writing code to traverse them and it is sufficiently extensible.

aviatesk: Sounds right. Thanks for your explanation.

clason:
Hi, Neovim maintainer here 👋 (and Julia user, hence my following this LS specifically).

We obviously approve of the client-agnostic approach, but would like to suggest going even further with that: Your first target should be the [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification) _exactly as written_  -- if that is unclear in any point, request clarification by opening an issue at the repo instead of assuming that VS Code represents "author intent". 

This is the official -- and hard -- stance of the Neovim project, and any off-spec features will be at best ignored and at worst break the feature on non-VS Code clients. (Such features belong in the VS Code extension, not the server.)

aviatesk:
Hello. Welcome for your comment. But I actually don't quite understand what you're actually proposing.

We read the specification carefully. That said, while LSP provides a framework for clients to inform servers about the markdown spec information that the client uses (`MarkdownClientCapabilities`), it's not obvious how clients handle markdown links. This is especially true when it comes to how LSP URIs are handled.
Reasonable client implementations would likely respect LSP URIs in their handling of markdown links, but there will be clients that don't. The realistic and appropriate approach that the server side can and should take is to provide functionality in a way that can be supported by as many clients as possible.

clason:
I am not proposing anything specific in this context; I just caught your comment

> This isn't fully compliant with LSP

and wanted to take the opportunity to raise the general point about not making _any_ client assumptions (so I am vehemently agreeing with you, it seems).

As long as the `hover` content is valid Markdown, Neovim will show it as such (and even make such links "clickable" via OSC8 -- what happens when you click on it depends on the terminal and the OS, of course).

aviatesk:
Yeah, the intent of my comment is rather to provide an implementation that would likely work in most cases for parts where the LSP doesn't specify the details of its behavior. And thank you for telling me about the specific implementation details of Neovim.

PR #154, opened 2025-07-04
add TestRunner.jl integration for inline test execution

aviatesk:
This commit introduces integration with TestRunner.jl to enable inline test execution directly from the editor. The implementation provides code lens `▶ Run` for `@testset` macros and as well as equivalent code actions for those clients that do not make much uses of code lenses, allowing users to run individual test sets and view their results without leaving the editor.

- Code lens/actions support for running individual `@testset` blocks - Implement `workspace/executeCommand` protocol for programmatic test execution using TEstRunner.jl - Test result diagnostics displayed inline with source code via `textDocument/diagnostics` - The test results are shown in updated code lenses where client/user make them visible, while also shown via `window/showMessage` with clickable actions of `▶ Rerun [summary]` to rerun it, `☰ Open logs` to show the full test outputs, and `✓ Clear result` to clear the results (test summary and diagnostics) - Temporary document creation for viewing test logs (for clients that support `window/showDocument`)


aviatesk:
@abap34 If you have a moment, could you test this PR on your client and let me know if you find any problems?

abap34:
Sure! I'll test it out.  (Personally, I think this feature is amazing 😳)

src/testrunner.jl
```diff
@@ -0,0 +1,489 @@
...
+    end
+    return code_actions
+end
+
+function check_testrunner_installed()
+    whichproc = open(`which testrunner`; read=true)
+    wait(whichproc)
+    testrunnerpath = String(read(whichproc.out))
+    return iszero(whichproc.exitcode) && !isempty(testrunnerpath)
+end
```
abap34: Maybe it would be better to use `Sys.which` ? It seems to make things work well across different OS.

abap34:
image

I tried it with both VSCode and Helix, and it worked smoothly! ☺️
Unfortunately, the display of result in Helix doesn't work, but it seems to be their issue: https://github.com/helix-editor/helix/issues/4699

aviatesk:
It's unfortunate that `window/showMessageRequest` is not supported.
I use `window/showMessageRequest` as a fallback for clients that don't support Code lens (both Helix and Zed fall into this category) when displaying test results and guiding users to open logs, but if that's also not available, we might be out of options...
Though we can get diagnostics, so we might at least be able to tell pass or fail at a minimum.

aviatesk:
FWIW, `Rerun "Test 1" [ Total: 3 | ... ]` is because of the limitation of TestRunner.jl. The integration with JL again is critical for TestRunner.jl to analyze precise location information of nested `@testset` clauses.. The workaround at this moment is insert arbitrary code construct just after the parent `@testset`, like
```julia
@testset "Test for hoge.jl" begin
    nothing
    @testset "Test 1" ...
    ...
end
```

src/testrunner.jl
aviatesk: Great suggestion, thanks.

abap34:
> It's unfortunate that `window/showMessageRequest` is not supported. I use `window/showMessageRequest` as a fallback for clients that don't support Code lens (both Helix and Zed fall into this category) when displaying test results and guiding users to open logs, but if that's also not available, we might be out of options... Though we can get diagnostics, so we might at least be able to tell pass or fail at a minimum.

Personally, I think it's okay to leave it later. The results can be checked from the Code Action, which is not that inconvenient for me.

It looks like there are some UI-related problems, but there's already a working implementation, maybe it won't take too long until it's supported, so it's not a bad idea to wait in anticipation of that. (https://github.com/helix-editor/helix/pull/5492)

PR #174, opened 2025-07-08
make `byte_ancestors` return empty list for out-of-range positions

aviatesk:
Initialize empty result containers and add range checks before adding root nodes. Also updates `select_target_node` accordingly. I think other places where `byte_ancestors` is used handle empty anscestor list case.

### DIFF
--- a/test/utils/test_ast.jl
+++ b/test/utils/test_ast.jl
@@ -53,6 +53,147 @@ end
     end
 end
 
+@testset "`byte_ancestors`" begin
+    # Test with a simple function
+    let code = """
+        function foo(x)
+            return x│ + 1
+        end
+        """
+        clean_code, positions = JETLS.get_text_and_positions(code, r"│")
+        return_pos = JETLS.xy_to_offset(Vector{UInt8}(clean_code), positions[1])
+
+        # Test with JL.SyntaxTree
+        st = jlparse(clean_code)
+
+        # Test at the beginning of "function"
+        ancestors = JETLS.byte_ancestors(st, 1)
+        @test length(ancestors) >= 2
+        @test JS.kind(ancestors[1]) === JS.K"function"
+        @test JS.kind(ancestors[end]) === JS.K"toplevel"
+
+        # Test after "return x"
+        ancestors = JETLS.byte_ancestors(st, return_pos)
+        @test length(ancestors) >= 4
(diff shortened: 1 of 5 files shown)


test/utils/test_ast.jl
```diff
@@ -53,6 +53,147 @@ end
...
+
+        # Test that we correctly find the identifier at the multi-byte position
+        st = jlparse(clean_code)
+        ancestors_st = JETLS.byte_ancestors(st, γ_pos)
+        @test_broken any(node -> JS.kind(node) === JS.K"Identifier" && JS.sourcetext(node) == "γ", ancestors_st)
+
+        sn = jsparse(clean_code)
+        ancestors_sn = JETLS.byte_ancestors(sn, γ_pos)
+        @test_broken any(node -> JS.kind(node) === JS.K"Identifier" && JS.sourcetext(node) == "γ", ancestors_sn)
+    end
```
aviatesk: @mlechu It's not urgent, but when you have time, could you investigate this issue?
It seems that the behavior of `xy_to_offset` changes depending on whether the text includes multi-byte characters or not. This test case and the case immediately before it should return the same result, but it appears that only this case is broken, probably due to an issue with the implementation of `xy_to_offset`.

IIUC `JETLS.get_text_and_positions` seems to be returning the correct result:
```julia
julia> Int(JETLS.get_text_and_positions("a = b + │γ", r"│")[2][1].character)
8

julia> Int(JETLS.get_text_and_positions("α = γ + │γ", r"│")[2][1].character)
10
```

mlechu: I think `get_text_and_positions` might be producing incorrect output.  `LSP.Position`'s column is supposed to be a character offset, not a byte offset (see [here](https://github.com/aviatesk/JETLS.jl/blob/caed8ecb4b45544110eed75ade43f4cb9aec0bb3/LSP/src/basic-json-structures.jl#L61-L64)).

`xy_to_offset` assumes a Position with a character offset, so it calls `nextind` [here](https://github.com/aviatesk/JETLS.jl/blob/caed8ecb4b45544110eed75ade43f4cb9aec0bb3/src/utils/ast.jl#L29) to convert character to byte offset.

aviatesk: I should have read the doc.. Thanks.

PR #177, opened 2025-07-11
Bump JuliaSyntax and JuliaLowering for macro support

mlechu:
Main change: ParseStream no longer holds a flat token list; we need to do
     something to get it.  Arguably we should do something smarter than seeking
     through the tokens but I've left that for later.

TODO: preparing a commit where I try to delete the macro hack

### DIFF
--- a/src/signature-help.jl
+++ b/src/signature-help.jl
@@ -300,16 +300,14 @@ Return the last byte at or before `b` that isn't in whitespace or a comment (or
 if `pass_newlines=false`, also isn't a newline).
 """
 function prev_nontrivia_byte(ps::JS.ParseStream, b::Int; pass_newlines=false)
-    ti = get_current_token_idx(ps, min(JS.last_byte(ps), b))
-    isnothing(ti) && return nothing
-    skip = (i::Int) -> let tok = ps.tokens[i]; k = kind(tok)
-        b < tok.next_byte || JS.is_whitespace(k) && (pass_newlines || k != K"NewlineWs")
+    tc = token_at_offset(ps, min(JS.last_byte(ps), b))
+    isnothing(tc) && return nothing
+    while (JS.is_whitespace(kind(this(tc))) && (pass_newlines || kind(this(tc)) != K"NewlineWs"))
+        b < tc.next_byte && return b # in the middle of a token
+        tc = prev_tok(tc)
+        isnothing(tc) && return nothing
     end
-    while skip(ti)
-        ti -= 1
-        ti < 1 && return nothing
-    end
-    return ps.tokens[ti].next_byte - 1
+    return Int(last_byte(tc))
 end
 
 function is_relevant_call(call::JL.SyntaxTree)
@@ -329,7 +327,7 @@ function call_is_decl(_bas::JL.SyntaxList, i::Int, _basᵢ::JL.SyntaxTree = _bas
     return j <= lastindex(_bas) &&
         kind(_bas[j]) in JS.KSet"macro function" &&
         # in `f(x) = g(x)`, return true in `f`, false in `g`
-        _bas[j - 1] === _bas[j][1]
+        _bas[j - 1]._id === _bas[j][1]._id
 end
 
 # Find cases where a macro call is not surrounded by parentheses
@@ -352,8 +350,8 @@ cursor would be descendents of it.
 """
 function cursor_call(ps::JS.ParseStream, st0::JL.SyntaxTree, b::Int)
     # disable signature help if invoked within comment scope
-    prev_token_idx = get_prev_token_idx(ps, b)
-    if !isnothing(prev_token_idx) && JS.kind(ps.tokens[prev_token_idx]) === K"Comment"
+    tc = token_before_offset(ps, b)
+    if !isnothing(tc) && JS.kind(this(tc)) === K"Comment"
         return nothing
     end
 
--- a/src/signature-help.jl
+++ b/src/signature-help.jl
@@ -327,7 +327,7 @@ function call_is_decl(_bas::JL.SyntaxList, i::Int, _basᵢ::JL.SyntaxTree = _bas
     return j <= lastindex(_bas) &&
         kind(_bas[j]) in JS.KSet"macro function" &&
         # in `f(x) = g(x)`, return true in `f`, false in `g`
-        _bas[j - 1]._id === _bas[j][1]._id
+        _bas[j - 1]._id == _bas[j][1]._id
 end
 
 # Find cases where a macro call is not surrounded by parentheses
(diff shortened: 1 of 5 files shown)


src/signature-help.jl
```diff
@@ -329,7 +327,7 @@ function call_is_decl(_bas::JL.SyntaxList, i::Int, _basᵢ::JL.SyntaxTree = _bas
     return j <= lastindex(_bas) &&
         kind(_bas[j]) in JS.KSet"macro function" &&
         # in `f(x) = g(x)`, return true in `f`, false in `g`
-        _bas[j - 1] === _bas[j][1]
+        _bas[j - 1]._id === _bas[j][1]._id
```
aviatesk: nitpick: For comparing numeric characters, it's generally better to use `==` since it can handle cases where the numeric values have different types but represent the same value, e.g. `0x01 == 1`. Of course, in this case I believe the `_id` field is probably always `::Int`, so `===` would also work fine.

src/utils/ast.jl
```diff
@@ -186,39 +186,67 @@ function remove_macrocalls(st0::JL.SyntaxTree)
...
+function token_at_offset(ps::JS.ParseStream, offset::Int)
+    for tc in TokenCursor(ps)
+        tc.next_byte > offset && return tc
+    end
+    return nothing
 end
 
-function get_current_token_idx(fi::FileInfo, pos::Position)
+function token_at_offset(fi::FileInfo, pos::Position)
     fi === nothing && return nothing
```
aviatesk:
```suggestion
```

src/utils/ast.jl
```diff
@@ -186,39 +186,67 @@ function remove_macrocalls(st0::JL.SyntaxTree)
...
+Similar to `token_at_offset`, but when you need token `a` from `a| b c`
 """
-function get_prev_token_idx(ps::JS.ParseStream, offset::Int)
-    get_current_token_idx(ps, offset - 1)
+function token_before_offset(ps::JS.ParseStream, offset::Int)
+    token_at_offset(ps, offset - 1)
 end
-function get_prev_token_idx(fi::FileInfo, pos::Position)
+function token_before_offset(fi::FileInfo, pos::Position)
     fi === nothing && return nothing
```
aviatesk:
```suggestion
```

aviatesk (approved):

src/utils/ast.jl
```diff
@@ -186,39 +186,65 @@ function remove_macrocalls(st0::JL.SyntaxTree)
...
+positions only.  To interpret this as a linear list of tokens in source order,
+just ignore any non-terminal node, and accumulate the byte lengths of terminals.
+"""
+struct TokenCursor
+    tokens::Vector{JS.RawGreenNode}
+    position::UInt32
+    next_byte::UInt32
+end
+TokenCursor(ps::JS.ParseStream) =
+    TokenCursor(filter(!JS.is_non_terminal, ps.output), 1, 1)
```
mlechu, line 202: Not sure if filtering is best or I should just skip over the ones we don't want in iteration (or if it matters).  ParseStream's .output is now something like 

```
julia> jsps("function foo(x, #=comment=#); x; #=comment2=# end").output
22-element Vector{JuliaSyntax.RawGreenNode}:
 TOMBSTONE (0 bytes, orig_kind=TOMBSTONE)
 function (8 bytes, orig_kind=function)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Identifier (3 bytes, orig_kind=Identifier)
 ( (1 bytes, orig_kind=()
 Identifier (1 bytes, orig_kind=Identifier)
 , (1 bytes, orig_kind=,)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Comment (11 bytes, orig_kind=Comment)
 ) (1 bytes, orig_kind=))
 call (19 bytes, 7 children)
 ; (1 bytes, orig_kind=;)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Identifier (1 bytes, orig_kind=Identifier)
 ; (1 bytes, orig_kind=;)
 block (4 bytes, 4 children)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Comment (12 bytes, orig_kind=Comment)
 Whitespace (1 bytes, orig_kind=Whitespace)
 end (3 bytes, orig_kind=end)
 function (49 bytes, 19 children)
 toplevel (49 bytes, 20 children)
```

where each element is a RawGreenNode, and the ones we're interested in are just the ones with no children.

aviatesk:
Thanks for working on this. Merge this whenever you're satisfied with the changes.

src/utils/ast.jl
aviatesk: What does `!is_non_terminal` remove here?
From the name `is_non_terminal`, it sounds like it's filtering things that don't have children, so does `!is_non_terminal` mean we iterate over things that have child elements?

mlechu: The double negation is a bit confusing.  terminal = leaf, so !!terminal = terminal = keeping leaves

aviatesk: I see, so you're asking whether `TokenCursor` iteration should recurse into tokens that contain child elements, or not right?
Currently, I don't think there are scenarios where we'd want to recurse, but they might arise in the future.
However, even in such cases, if we can create a separate `TokenCursor` for tokens of a parent token obtained through `TokenCursor` iteration over the parent tree and iterate over that, I feel that would be a better design and probably faster for cases when we don't care about child tokens (alternatively, we could add a new option to `TokenCursor` for recursion).
In any case, I think we can consider this when such cases actually occur.
Is this understanding correct?

mlechu: I think your understanding is correct.  Removing nodes with children was the simplest way to restore the linear token list we had before, so we're ignoring the tree structure for now:
```
julia> filter(!JS.is_non_terminal, jsps("function foo(x, #=comment=#); x; #=comment2=# end").output)
18-element Vector{JuliaSyntax.RawGreenNode}:
 TOMBSTONE (0 bytes, orig_kind=TOMBSTONE)
 function (8 bytes, orig_kind=function)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Identifier (3 bytes, orig_kind=Identifier)
 ( (1 bytes, orig_kind=()
 Identifier (1 bytes, orig_kind=Identifier)
 , (1 bytes, orig_kind=,)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Comment (11 bytes, orig_kind=Comment)
 ) (1 bytes, orig_kind=))
 ; (1 bytes, orig_kind=;)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Identifier (1 bytes, orig_kind=Identifier)
 ; (1 bytes, orig_kind=;)
 Whitespace (1 bytes, orig_kind=Whitespace)
 Comment (12 bytes, orig_kind=Comment)
 Whitespace (1 bytes, orig_kind=Whitespace)
 end (3 bytes, orig_kind=end)

```

Note there is also some tree-cursor iteration functionality in JuliaSyntax for actually treating the output as a tree.  If there's an easy JuliaSyntax way to iterate through all terminals, I didn't find it.

aviatesk: Ok, thanks for the clarification.

PR #178, opened 2025-07-11
TokenCursor: add more `iterator` interface definitions

aviatesk:
To make `collect(tc::TokenCursor)` etc. available, which are useful for interactive uses.


mlechu (approved):
👍

Not a priority now, but the printing for `RawGreenNode`s could also be improved for interactive use

aviatesk:
Yeah, I agree. We should improve that also.

PR #179, opened 2025-07-11
ast: add tests, add `next_nontrivia`, fix `prev_nontrivia`

aviatesk:


test/utils/test_ast.jl
```diff
@@ -309,6 +309,177 @@ end
...
+        @test String(ps.textbuf[JETLS.byte_range(@something(JETLS.token_at_offset(ps, 6)))]) == "bar"
+    end
+
+    # Test token_before_offset
+    let code = "a + b"
+        ps = parsedstream(code)
+        @test String(ps.textbuf[JETLS.byte_range(@something(JETLS.token_before_offset(ps, 2)))]) == "a"
+        @test String(ps.textbuf[JETLS.byte_range(@something(JETLS.token_before_offset(ps, 3)))]) == " "
+        @test String(ps.textbuf[JETLS.byte_range(@something(JETLS.token_before_offset(ps, 4)))]) == "+"
+        @test JETLS.token_before_offset(ps, 1) !== nothing  # COMBAK should probably return nothing instead
```
aviatesk: I think we should remove `TOMBSTONE` from the token list that `TokenCursor` returns.
In this case, returning `nothing` seems more natural, and we should also be able to remove the special case for `prev_nontrivia`.

src/signature-help.jl
```diff
@@ -375,10 +360,8 @@ function cursor_call(ps::JS.ParseStream, st0::JL.SyntaxTree, b::Int)
     end
 
     # `i` is nothing.  Eat preceding whitespace and check again.
-    pnb_line = prev_nontrivia_byte(ps, b; pass_newlines=false)
-    pnb = prev_nontrivia_byte(ps, b; pass_newlines=true)
-
-    (isnothing(pnb) || pnb === b) && return nothing
+    pnb = prev_nontrivia_byte(ps, b-1; pass_newlines=true)
```
aviatesk, line 363: `-1` because now we have `1 ≤ b ≤ JS.last_byte(ps) || return nothing` check in `prev_nontrivia`, but I believe giving `b-1` here is more correct?

PR #180, opened 2025-07-12
Allow macros through JuliaLowering

mlechu:
Macros now expand!  

This PR:
- Expr-macros generally work through JuliaLowering and ST->expr->ST conversion
- Macro calls defer to the new-style macro method if available (for provenance, and because the fixed number of new-style macros we have produce weird forms that are not necessarily converted well)
- Macro definitions create expr-macros in compat mode (default on), so we shouldn't get new macros where we don't want them.
- In the current implementation, every node from an expansion gets provenance equal to the macrocall that caused it.  This might change to use the linenodes we collected in expr->ST conversion instead.

TODO:
- Test more common macros (e.g. `@time` fails in evaluation, but might be a lowering bug)
- Hygiene and escaping is currently unhandled
- Write more syntax_macros.jl-type macros in ST form to preserve provenance


test/test_completions.jl
```diff
@@ -269,6 +270,12 @@ function with_completion_request(tester::Function, text::AbstractString; matcher
 end
 
 @testset "get_completion_items" begin
+    Base.include_string(lowering_module, """
+        macro weirdmacro(x::Symbol, v)
+            name = Symbol(string(x, "_var"))
+            return :(\$(esc(name)) = \$v)
+        end""")
+
```
mlechu: Hacky, but not sure how else to get the macro definition into the module.  Ideally the LS should know about the definition.

test/test_hover.jl
```diff
@@ -166,10 +166,10 @@ end
         # M_doc.func│(42)
         (; pat="Documented method.")
 
-        # sinx = @inline│ sin(42)
+        # sinx = @inlin│e sin(42)
         (; pat=JETLS.lsrender(@doc @inline))
 
-        # sinx = Base.@inline│ sin(42)
+        # sinx = Base.@inlin│e sin(42)
```
mlechu: Could add a test_broken to the one at the end of the identifier, when the cursor isn't in the macroname node.

src/utils/binding.jl
```diff
@@ -16,15 +16,19 @@ function is_relevant(ctx::JL.AbstractLoweringContext,
          || cursor > start)
 end
 
-let lowering_module = Module()
-    global function jl_lower_for_scope_resolution2(st0)
-        ctx1, st1 = JL.expand_forms_1(lowering_module, remove_macrocalls(st0));
+# Lowering doesn't really require the module apart from looking up macro names.
+# If a feature takes a LoweringModule, most functionality will be present with
+# mod=nothing.
+const LoweringModule = Union{Module, Nothing}
```
aviatesk:
```suggestion
const MaybeLoweringModule = Union{Module, Nothing}
```

src/utils/binding.jl
```diff
@@ -134,7 +138,7 @@ function select_target_binding(ctx3::JL.VariableAnalysisContext, st3::JL.SyntaxT
 end
 
 """
-    select_target_binding_definitions(st0_top::JL.SyntaxTree, offset::Int) ->
+    select_target_binding_definitions(st0_top::JL.SyntaxTree, offset::Int[, mod]) ->
```
aviatesk, line 141:
```suggestion
    select_target_binding_definitions(st0_top::JL.SyntaxTree, offset::Int, mod::MaybeLoweringModule) ->
```

test/test_completions.jl
```diff
@@ -10,9 +10,10 @@ using JETLS.URIs2
 include("setup.jl")
 include("jsjl_utils.jl")
 
+lowering_module = Module()
```
aviatesk:
```suggestion
global lowering_module::Module = Module()
```

test/test_completions.jl
aviatesk: The macro definitions contained in user code should be loaded (currently by the native lowerer) through the full analysis performed inside `with_completion_request`, so it should work without this hack, but would removing this cause problems?

test/test_completions.jl
```diff
@@ -327,7 +334,7 @@ end
             @test any(items) do item
                 item.label == "x"
             end
-            @test_broken !any(items) do item
+            @test !any(items) do item
```
aviatesk, line 331: 🎉

Project.toml
```diff
@@ -21,7 +21,7 @@ REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 [sources]
 JET = {rev = "master", url = "https://github.com/aviatesk/JET.jl"}
 JSONRPC = {path = "JSONRPC"}
-JuliaLowering = {rev = "jetls-hacking", url = "https://github.com/mlechu/JuliaLowering.jl"}
+JuliaLowering = {rev = "302fe43", url = "https://github.com/mlechu/JuliaLowering.jl"}
```
aviatesk: Commits that have not yet been merged into `master` may be lost due to force push, so it's preferable to specify the remote branch name. Of course, remote branches can also always be deleted though.

mlechu: Makes sense, I just didn't want to accidentally change JETLS behaviour by pointing it at a moving target.  Pick whichever you prefer.

test/test_hover.jl
mlechu: Actually, I need to see the UI for this in different editors.  `@inline|` is hovering over the space after `@inline` with a mouse, I think

test/test_completions.jl
mlechu: Yup, it works without (not sure why it wasn't working before).

Project.toml
mlechu: OK, pinning the rev makes it hard to make tweaks to `fix-nightly` or `expr-to-st`.  I'll use the branch name and be careful about breaking stuff.

aviatesk:
> Hygiene and escaping is currently unhandled
> 
>     * Namely, every macro expansion gets its own scope layer right now with no escaping, but JETLS doesn't look at the scope layers, so we just have unhygienic macros.  For things like completions, this is OK (though maybe not for binding jump).

How difficult would it be to support this?
The unused variable detection in particular should suffer from this.
For example, code like the following:
```julia
julia> st0 = """
       let s = rand(Int)
           lazy"s = \$s"
       end
       """ |> jlparse
SyntaxTree with attributes kind,value,name_val,syntax_flags,source
[toplevel]                                                   │
  [let]                                                      │
    [block]                                                  │
      [=]                                                    │
        s                                :: Identifier       │
        [call]                                               │
          rand                           :: Identifier       │
          Int                            :: Identifier       │
    [block]                                                  │
      [macrocall]                                            │
        nothing                          :: StringMacroName  │
        [string-r]                                           │
          "s = \$s"                      :: String           │


julia> ctx3, st3 = JETLS.jl_lower_for_scope_resolution3(st0[1], JETLS);

julia> st3
SyntaxTree with attributes kind,var_id,scope_type,lambda_bindings,name_val,syntax_flags,meta,scope_layer,value,is_toplevel_thunk,source
[lambda]                                                     │ lambda_bindings,is_toplevel_thunk=true
  [block]                                                    │
  [block]                                                    │
  [block]                                                    │
    [=]                                                      │
      #₃/tmp                             :: BindingId        │
      [call]                                                 │
        #₅/rand                          :: BindingId        │
        #₄/Int                           :: BindingId        │
    [block]                                                  │
      [local]                                                │
        #₆/s                             :: BindingId        │
      nothing                            :: TOMBSTONE        │
      nothing                            :: TOMBSTONE        │
      [=]                                                    │
        #₆/s                             :: BindingId        │
        #₃/tmp                           :: BindingId        │
      [block]                                                │
        [call]                                               │
          #₇/LazyString                  :: BindingId        │
          "s = "                         :: String           │
          #₈/s                           :: BindingId        │
```

Due to `#₆/s != #₈/s`, `#₆/s` will be reported as an unused variable, and when passing `CodeInfo` generated by `JL.to_lowered_expr` to `LSAnalyzer`, it will be reported as an undefined global variable.

aviatesk:
Also, it seems that the macro definitions defined in jl_syntax_macros.jl are only used when `lowering_module===JETLS`, but why is this the case?
Since they are defined with `Base.var"@nospecialize"(...)`, I expected them to be available globally, but it appears that's not the case.

```julia
julia> st0 = """
       kwargs_dict(@nospecialize configs) = nothing
       """ |> jlparse;

julia> _, st31 = JETLS.jl_lower_for_scope_resolution3(st0[1]);

julia> _, st32 = JETLS.jl_lower_for_scope_resolution3(st0[1], JETLS);

julia> configs1 = st31[3][3][1][2][1][2][3][1][2]
SyntaxTree with attributes kind,var_id,scope_type,lambda_bindings,name_val,syntax_flags,meta,scope_layer,value,is_toplevel_thunk,source
#₆/configs                               :: BindingId        │

julia> configs2 = st32[3][3][1][2][1][2][3][1][2]
SyntaxTree with attributes kind,var_id,scope_type,lambda_bindings,name_val,syntax_flags,meta,scope_layer,value,is_toplevel_thunk,source
#₆/configs                               :: BindingId        │

julia> JS.first_byte(configs1)
0

julia> JS.first_byte(configs2)
27
```

Project.toml
aviatesk: Yep I agree. We need to take care of compatiblity ourselves until JL is registered as an official package anyway.

mlechu:
> I expected them to be available globally, but it appears that's not the case.

Fixed by https://github.com/mlechu/JuliaLowering.jl/pull/1/commits/db49bd35462fa46ba9fa8e036039d8cae85d8a12

mlechu:
> How difficult would it be to support this?

I think @c42f was looking at this, but I could try to throw something together quickly on my branch.  To my understanding, esc(x) should be manageable, but JL would need modification to support esc(esc(x)).

The parts of JETLS that use `st0` would have to separately begin considering scope layers (I realize we usually use bindings, so maybe no changes needed).

aviatesk:
That's right, it would be helpful if you could support just `esc(x)` for now.
That should probably cover around 90% of the real world codebases. It would be sufficient coverage for us to start a proper integration with JL I think.

aviatesk:
Let's merge this once the basic macro hygine support has been added to the branch.

test/test_surface_analysis.jl
```diff
@@ -183,14 +183,11 @@ end
...
-            diagnostic = only(diagnostics)
-            @test diagnostic.message == "Unused local binding `configs`"
-            @test diagnostic.range.start.line == 1
-            @test diagnostic.range.var"end".line == 1
-        end
+        @test length(diagnostics) == 1
+        diagnostic = only(diagnostics)
+        @test diagnostic.message == "Unused argument `configs`"
+        @test diagnostic.range.start.line == 0
+        @test diagnostic.range.var"end".line == 0
```
aviatesk, line 190: 👍

test/test_hover.jl
aviatesk: I reverted the change since it's working correctly as is. JFYI, you can run the test just with `julia --startup-file=no test/test_hover.jl` without using the full client UI.

aviatesk (approved):
Thanks so much for your work on this, @mlechu

PR #185, opened 2025-07-16
implement basic configuration system

abap34:
config_demo_v1

This PR introduces a basic configuration system.

Currently, the implementation simply reads from `$(rootPath)/JETLSConfig.toml` at startup and watches the file using `workspace/didChangeWatchedFiles`.

In the future, we plan to support additional sources such as:

* A global configuration file (e.g., `~/.config/JETLS/JETLSConfig.toml`)
* Reading from `Project.toml`
* Command-line arguments
* `workspace/configuration` requests

The configuration system classifies all settings into two categories:

1. Settings that can be updated dynamically.
2. Settings that require a restart to take effect.

For the first category, updates are directly applied to `config_manager.actual_config`. For the second, the user is notified via `window/showMessage`, but the values themselves are not updated — this is purely to reduce unnecessary notifications.


src/utils/config.jl
```diff
@@ -0,0 +1,129 @@
...
+    "performance" => Dict{String,Any}(
+        "full_analysis" => Dict{String,Any}(
+            "debounce" => true,
+            "throttle" => true
+        )
+    ),
+    "testrunner" => Dict{String,Any}(
+        "executable" => false,
+    ),
+)
```
abap34: A quick investigation suggests that registering our configuration schema on platforms like https://www.schemastore.org/ could enable features such as autocompletion and validation for users easily and widely. There also appear to be various easy editor-specific mechanisms for supporting. Of course, this is independent of the language server itself.

However, this investigation is still incomplete, so the default configuration values are hardcoded for now. As the number of configuration options grows, we will likely move both the values and the schema to an external configuration file, after deciding on a more detailed approach.

src/analysis/full-analysis.jl
```diff
@@ -26,13 +22,13 @@ function run_full_analysis!(server::Server, uri::URI; onsave::Bool=false, token:
             end
             id = hash(run_full_analysis!, hash(analysis_unit))
             if onsave
-                debounce(id, FULL_ANALYSIS_DEBOUNCE) do
-                    throttle(id, FULL_ANALYSIS_THROTTLE) do
+                debounce(id, get_config(server.state.config_manager, ["performance", "full_analysis", "debounce"])) do
```
aviatesk: We should probably change the interface like `get_config(server, "performance", "full_analysis", "debounce")` letting inference unroll internal recursion while avoiding allocating arrays.

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,93 @@
...
+const DID_CHANGE_WATCHED_FILES_REGISTRATION_METHOD = "workspace/didChangeWatchedFiles"
+
+function did_change_watched_files_registration()
+    Registration(;
+        id = DID_CHANGE_WATCHED_FILES_REGISTRATION_ID,
+        method = DID_CHANGE_WATCHED_FILES_REGISTRATION_METHOD,
+        registerOptions = DidChangeWatchedFilesRegistrationOptions(;
+            watchers = [
+                FileSystemWatcher(;
+                    globPattern = "**/JETLSConfig.toml",
```
aviatesk: How about renaming `JETLSConfig.toml` to `.JETLSConfig.toml`? I don't think there is a standard of how to name configurations files in our ecosystem, but at least we have `.JuliaFormatter.toml`.

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,93 @@
...
+# For dynamic registrations during development
+# unregister(currently_running, Unregistration(;
+#     id=DID_CHANGE_WATCHED_FILES_REGISTRATION_ID,
+#     method=DID_CHANGE_WATCHED_FILES_REGISTRATION_METHOD))
+# register(currently_running, did_change_watched_files_registration())
+
+function initialize_config!(server::Server)
+    if !isdefined(server.state, :root_path)
+        if JETLS_DEV_MODE
+            @info "server.state.root_path is not defined, skipping add JETLSConfig.toml watcher"
```
aviatesk:
```suggestion
            @info "`server.state.root_path` is not defined, skipping registering JETLSConfig.toml watcher"
```

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,93 @@
...
+        end
+        return
+    end
+    config_path = joinpath(server.state.root_path, "JETLSConfig.toml")
+    push!(server.state.config_manager.watching_files, config_path)
+    load_config!(server, config_path) do actual_config, latest_config, key_path, v
```
aviatesk:
```suggestion
    load_config!(server, config_path) do actual_config, latest_config, key, v
```
Maybe we can make `merge_config!` pass `key` only for the callback.

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,93 @@
...
+function load_config!(on_reload_required, server::Server, path::AbstractString)
+    isfile(path) || return
+    parsed = TOML.tryparsefile(path)
+    parsed isa TOML.ParserError && return # just skip to reduce noise while typing
+    unknown_keys = collect_unmatched_keys(parsed)
+
+    if !isempty(unknown_keys)
+        show_error_message(server, """
+            Configuration file at $path contains unknown keys.
+            unknown keys: $(join(map(x -> join(x, "."), unknown_keys), ", "))
```
aviatesk:
```suggestion
            Configuration file at $path contains unknown keys:
            $(join(map(x -> join(x, "."), unknown_keys), ", "))
```

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,93 @@
...
+                    Configuration key `$(join(key_path, "."))` was changed.
+                    Please restart the server to apply the changes.
+                    """)
+            end
+        end
+    elseif change_type == FileChangeType.Deleted
+        show_warning_message(server, """
+            $changed_path was deleted.
+            Please restart the server to apply the changes.
+            """)
```
aviatesk: How about extending the `ConfigManager` data structure to maintain a mapping of `config_file_path => configs`?
That way, when a configuration file is deleted, it should be possible to delete the configuration held in that mapping and revert to the default.
In the future, when we want to look at multiple configuration files (such as those under `~/.config`), such a mapping will be necessary to implement a configuration that reflects that hierarchy.

src/testrunner.jl
```diff
@@ -435,13 +437,15 @@ end
...
 function testrunner_run_testcase(server::Server, uri::URI, tcl::Int, tct::String, filepath::String;
                                  token::Union{Nothing,ProgressToken}=nothing)
-    if isnothing(Sys.which("testrunner"))
+    testrunner_exe_name = get_config(server.state.config_manager, ["testrunner", "executable"])
+    if isnothing(Sys.which(testrunner_exe_name))
         show_error_message(server, """
-            `testrunner` executable is not found on the `PATH`.
-            Follow [this instruction](https://github.com/aviatesk/TestRunner.jl#installation)
-            to install the `testrunner` app.
+            `$testrunner_exe_name` executable is not found on the `PATH``.
```
aviatesk:
```suggestion
            `$testrunner_exe_name` executable is not found on the `PATH`.
```

src/types.jl
```diff
@@ -218,6 +218,14 @@ struct LSPostProcessor
 end
 LSPostProcessor() = LSPostProcessor(JET.PostProcessor())
 
+struct ConfigManager
+    actual_config::Dict{String,Any} # Configuration that actually used by the server
+    latest_config::Dict{String,Any} # Latest configuration written to the file to suppress excessive messages
+    watching_files::Set{String}
```
aviatesk:
```suggestion
    watched_files::Set{String}
```

test/utils/test_config.jl
```diff
@@ -0,0 +1,104 @@
...
+        "nested_key1" => false,
+        "nested_key2" => Dict(
+            "deep_nested_key1" => true,
+            "deep_nested_key2" => false
+        )
+    )
+)
+
+JETLS.is_reload_required_key(key_path::Vector{String}) =
+    JETLS.access_nested_dict(TEST_RELOAD_REQUIRED, key_path)
```
aviatesk: Such overrides can cause serious invalidation, so they should be avoided as much as possible, even in tests.
```julia
global CONFIG_RELOAD_REQUIRED::Dict{String,Any} = ...
```
I think it would be better to set it up like this, and then wrap the test side with `try/finally` while replacing it with `JETLS.CONFIG_RELOAD_REQUIRED = ...`.

test/test_did-change-watched-files.jl
```diff
@@ -0,0 +1,179 @@
...
+            @test occursin("restart", raw_res.params.message)
+            @test server.state.config_manager.actual_config["performance"]["full_analysis"]["debounce"] == JETLS.DEFAULT_CONFIG["performance"]["full_analysis"]["debounce"]
+            @test server.state.config_manager.latest_config["performance"]["full_analysis"]["debounce"] == 200.0
+            @test server.state.config_manager.actual_config["testrunner"]["executable"] == "mynewtest"
+            @test server.state.config_manager.latest_config["testrunner"]["executable"] == "mynewtest"
+        end
+    end
+end
+
+
```
aviatesk:
```suggestion
```

src/analysis/full-analysis.jl
abap34: I agree with you about the interface.

As for the debounce/throttle, this implementation was from original implementation.
I will change it if it wasn’t intentional.

https://github.com/aviatesk/JETLS.jl/blob/266faff1d6e4b7cca5c0641a5f9779f2e7cdc8f7/src/analysis/full-analysis.jl#L29-L30

src/did-change-watched-files.jl
abap34: It seems better for me. I'll make the change :+1:

src/analysis/full-analysis.jl
aviatesk: I suggested the change on `get_config` only. I'm welcome to revisit the debounce/throttle logic used here, but it should probably be done in another PR.

src/did-change-watched-files.jl
abap34: Actually, We need full `key_path` here.

https://github.com/aviatesk/JETLS.jl/blob/34df7c63498da803d94b2b586a29c444ad133ce6/src/did-change-watched-files.jl#L71-L80

src/did-change-watched-files.jl
abap34: I agree that your suggested implementation would likely be more robust to future changes. I’ve updated the code accordingly, though I haven’t yet fully finished the implementation details. I’ll let you know once it’s ready for a review.

- Each config file’s contents are now stored in a `SortedDict` mapping filename => content. Default settings and command-line arguments are also stored in this dict using special reserved names.
- The dict is sorted by `ConfigFileOrder`, so config values can be retrieved correctly by iterating it.
- Removing a file simply handled deleting its corresponding key from the dict.

test/test_did-change-watched-files.jl
```diff
@@ -0,0 +1,177 @@
...
+        rootUri = filepath2uri(tmpdir)
+        withserver(; rootUri, capabilities=CLIENT_CAPABILITIES) do (; writereadmsg, readmsg, id_counter, server)
+            @test server.state.config_manager.actual_config["performance"]["full_analysis"]["debounce"] == 2.0
+            @test server.state.config_manager.latest_config["performance"]["full_analysis"]["debounce"] == 2.0
+
+            @test server.state.config_manager.actual_config["testrunner"]["executable"] == "mytestrunner"
+            @test server.state.config_manager.latest_config["testrunner"]["executable"] == "mytestrunner"
+
+            open(config_path, "w") do io
+                write(io, "[performance.full_analysis]\ndebounce = 3.0\n[testrunner]\nexecutable = \"mytestrunner\"\n")
```
aviatesk:
```suggestion
                write(io, """
                [performance.full_analysis]
                debounce = 3.0
                
                [testrunner]
                executable = "mytestrunner"
                """)
```
may be easier to read?

src/utils/config.jl
```diff
@@ -0,0 +1,201 @@
...
+
+function access_nested_dict(dict::AbstractDict, path::String, rest_path::String...)
+    nextobj = @something get(dict, path, nothing) return nothing
+    return access_nested_dict(nextobj, rest_path...)
+end
+access_nested_dict(obj) = obj
```
aviatesk:
```suggestion
function access_nested_dict(dict::Dict{String,Any}, path::String, rest_path::String...)
    nextobj = @something get(dict, path, nothing) return nothing
    nextobj isa Dict{String,Any} || return nextobj
    return access_nested_dict(nextobj, rest_path...)
end
```

abap34:
This PR is now ready for review. Several parts of the implementation have been updated from last review.
 Here’s a brief summary of the changes:

1. `ConfigManager` now retains the configuration entries for each config file individually.
2. On startup, all config files are merged according to their priority. Keys that require a reload have their values fixed and stored separately. Access to these keys is now routed through this fixed dict to ensure consistency.
3. Other keys can still be updated as before. However, unlike the previous behavior, config lookup now follows the priority order of files, and the first matching value is returned.
4. `ConfigManager` now manages the fixed values for reload-required keys, so `actual_config` and `latest_config` have been removed.

src/utils/config.jl
```diff
@@ -0,0 +1,257 @@
...
+                         source::Dict{String,Any}, key_path::Vector{String})
+    for (k, v) in source
+        current_path = [key_path; k]
+        if v isa Dict{String,Any}
+            if haskey(target, k) && target[k] isa Dict{String,Any}
+                traverse_merge!(on_leaf, target[k], v, current_path)
+            else
+                target[k] = Dict{String,Any}()
+                traverse_merge!(on_leaf, target[k], v, current_path)
+            end
```
aviatesk:
```suggestion
            tv = get(target, k, nothing)
            if tv isa Dict{String,Any}
                traverse_merge!(on_leaf, tv, v, current_path)
            else
                tv = target[k] = Dict{String,Any}()
                traverse_merge!(on_leaf, tv, v, current_path)
            end
```

test/test_did-change-watched-files.jl
```diff
@@ -0,0 +1,310 @@
...
+            # change `performance.full_analysis.debounce` to `DEBOUNCE_V2`
+            DEBOUNCE_V2 = 200.0
+            open(config_path, "w") do io
+                write(io, """
+                    [performance.full_analysis]
+                    debounce = $DEBOUNCE_V2
+                    [testrunner]
+                    executable = \"$TESTRUNNER_STARTUP\"
+                    """)
+            end
```
aviatesk:
```suggestion
            write(config_path, """
                [performance.full_analysis]
                debounce = $DEBOUNCE_V2
                [testrunner]
                executable = \"$TESTRUNNER_STARTUP\"
                """)
```

aviatesk:
Can you resolve the conflicts? Let me know if you have any troubles with it.

abap34:
The conflicts have been resolved.
The error message in testrunner.jl, which had significant conflicts, has been unified as follows:
- If a customized executable name for `testrunner` is set, a message prompting the user to check that setting is displayed.
- Otherwise, the traditional message prompting installation is shown.

I’m not strongly attached to this change, so please feel free to share any thoughts or suggestions you may have.

src/did-change-watched-files.jl
```diff
@@ -0,0 +1,109 @@
...
+
+const DID_CHANGE_WATCHED_FILES_REGISTRATION_ID = "jetls-did-change-watched-files"
+const DID_CHANGE_WATCHED_FILES_REGISTRATION_METHOD = "workspace/didChangeWatchedFiles"
+
+function did_change_watched_files_registration()
+    Registration(;
+        id = DID_CHANGE_WATCHED_FILES_REGISTRATION_ID,
+        method = DID_CHANGE_WATCHED_FILES_REGISTRATION_METHOD,
+        registerOptions = DidChangeWatchedFilesRegistrationOptions(;
+            watchers = [
```
aviatesk:
```suggestion
            watchers = FileSystemWatcher[
```

src/utils/general.jl
```diff
@@ -198,20 +198,34 @@ end
...
-function app_notfound_message(app::AbstractString)
+
+function check_settings_message(setting_path::String...)
+    """
+    This value is configured in the `$(join(setting_path, "."))`.
+    Please check the settings.
+    """
+end
+
+function app_notfound_message(app::AbstractString, is_default_setting::Bool=true, setting_path::String...)
```
aviatesk: Wow, I didn't know this can be a valid method signature.
```suggestion
function app_notfound_message(app::AbstractString, setting_path::String...; is_default_setting::Bool=true)
```
seems to be more straightforward?

src/utils/config.jl
```diff
@@ -0,0 +1,258 @@
...
+    for (k, v) in source
+        current_path = [key_path; k]
+        if v isa Dict{String,Any}
+            tv = get(target, k, nothing)
+            if tv isa Dict{String,Any}
+                traverse_merge!(on_leaf, tv, v, current_path)
+            else
+                tv = target[k] = Dict{String,Any}()
+                traverse_merge!(on_leaf, tv, v, current_path)
+            end
```
aviatesk, line 63:
```suggestion
            tv = get!(Dict{String,Any}, target, k)
            traverse_merge!(on_leaf, tv, v, current_path)
```

aviatesk (commented):
I think this is implemented very well overall. We could merge this as is, but in the future, I'd like to work on the following items also:
- Adding documentation for this configuration system
- Introducing configuration for diagnostics
- Improving multi-file handling

let me explain the last point in more detail.

Regarding configuration across multiple files, I was also envisioning support for cases like this:

> With `rootPath == "~/julia/packages/MyPkg.jl"`
```
- ~/julia/packages/MyPkg.jl/.JETLSConfig.toml
- ~/.JETLSConfig.toml
- __DEFAULT_CONFIG__
```

Here, `Base.lt(::ConfigFileOrder, path1, path2)` would be defined based on the directory hierarchy relationship between `path1` and `path2`, with subdir given higher priority, in addition to giving the current `__DEFAULT_CONFIG__` the lowest priority.

However, I don't think it's possible to watch `.JETLSConfig.toml` outside of `rootPath`, which could make it confusing which configuration change can be applied dynamically. So it might be better to only apply configuration from files under `rootPath` anyway.

On the other hand, in any case, I would like to support cases like this:
```
rootPath
├─ .JETLSConfig.toml (1)
├─ code.jl
├─ subdir1
│  ├─ .JETLSConfig.toml (2)
│  └─ sub_code1.jl
└─ subdir2
   └─ sub_code2.jl
```
Here, in this project, the settings from `.JETLSConfig.toml (1)` would be used, but for code under `subdir1`, I want to use `.JETLSConfig.toml (2)`. Especially when introducing diagnostics settings, such file hierarchy-aware configuration would be useful.
We probably need to change the implementation of `Base.lt(::ConfigFileOrder, path1, path2`, as well as change the signature of `get_config(manager::ConfigManager, key_path::String...)` to `get_config(manager::ConfigManager, curdir::String, key_path::String...)`.

---

Having said a lot,  since this PR itself is already quite a large change, it would be perfectly fine to tackle the above TODOs incrementally while merging this PR as is.
Please let me know how you'd like to proceed.

Project.toml
```diff
@@ -26,6 +28,7 @@ JuliaSyntax = {rev = "e02f29f", url = "https://github.com/JuliaLang/JuliaSyntax.
 LSP = {path = "LSP"}
 
 [compat]
+DataStructures = "0.18.22"
```
aviatesk:
```suggestion
DataStructures = "0.19"
```
The latest version of DataStructures should get rid of unnecessary dependency on Compat.jl.

src/utils/config.jl
abap34: There are cases where `tv` is not a `Dict` (like `String`), and even in those cases, we still need to create an empty dict. So I think the current implementation is the most straightforward.

abap34:
I’ve taken note of the points you mentioned, especially the configuration across multiple files.

As you pointed out, this PR is already quite large, and since there are other implementations (like recursive analysis) that I'm working and depend on this feature, I prefer to merge it now and address those items afterward.

aviatesk:
Ok, let's merge this as is then. Great work on this, @abap34 !

PR #187, opened 2025-07-16
Add JETLS_DEBUG_LOWERING for JL-related logging

mlechu:
Lowering failures shouldn't interfere with the LS running (or even LS development), but they are usually of interest for JL development, where I often end up with a bunch of local logging code.  This change puts lowering backtraces behind another Preferences.jl variable so adding JL-specific logging is less intrusive.

### DIFF
--- a/test/jsjl_utils.jl
+++ b/test/jsjl_utils.jl
@@ -14,6 +14,16 @@ jlparse(s::AbstractString; rule::Symbol=:all, kwargs...) = jlparse(parsedstream(
 jlparse(parsed_stream::JS.ParseStream; filename::AbstractString=@__FILE__, first_line::Int=1) =
     JS.build_tree(JL.SyntaxTree, parsed_stream; filename, first_line)
 
+macro expect_jl_err(ex)
+    if JETLS.JETLS_DEBUG_LOWERING
+        esc(:(redirect_stdio(stderr=>devnull) do
+                  $ex
+              end))
+    else
+        esc(ex)
+    end
+end
+
 # For interactive use
 # ===================
 
--- a/src/JETLS.jl
+++ b/src/JETLS.jl
@@ -11,8 +11,9 @@ end
 using Preferences: Preferences
 const JETLS_DEV_MODE = Preferences.@load_preference("JETLS_DEV_MODE", false)
 const JETLS_TEST_MODE = Preferences.@load_preference("JETLS_TEST_MODE", false)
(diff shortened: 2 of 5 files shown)


test/jsjl_utils.jl
```diff
@@ -14,6 +14,16 @@ jlparse(s::AbstractString; rule::Symbol=:all, kwargs...) = jlparse(parsedstream(
...
 
+macro expect_jl_err(ex)
+    if JETLS.JETLS_DEBUG_LOWERING
+        esc(:(redirect_stdio(stderr=>devnull) do
+                  $ex
+              end))
+    else
+        esc(ex)
+    end
+end
```
aviatesk:
```suggestion
macro expect_jl_err(ex)
    if JETLS.JETLS_DEBUG_LOWERING
        :(mktemp() do path, io
            redirect_stderr(io) do
                $ex
            end
            flush(io)
            @test !isempty(read(path, String))
        end) |> esc
    else
        esc(ex)
    end
end
```

aviatesk (approved):
Maybe we want to add this entry to .github/workflows/ci.yml too:

```yaml
  JETLS_DEBUG_LOWERING:
    name: Test with `JETLS_DEBUG_LOWERING = true`
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v1
        with:
          version: "1.12-nightly" # most stable version
          arch: x64
      - uses: julia-actions/cache@v2
      - name: set LocalPreferences.toml
        working-directory: .
        run: |
          echo '
          [JETLS]
          JETLS_DEBUG_LOWERING = true

          [JET] # allow JET to be loaded on nightly
          JET_DEV_MODE = true
          ' > LocalPreferences.toml
      - uses: julia-actions/julia-buildpkg@latest
      - uses: julia-actions/julia-runtest@latest
      - uses: julia-actions/julia-processcoverage@v1
      - uses: codecov/codecov-action@v5
```

mlechu:
> Maybe we want to add this entry to .github/workflows/ci.yml too

Feel free to add it if you like.  I'll probably be the only one using the flag anyway (unless you're submitting a bug report or something), so I'm also OK keeping the `JETLS_DEBUG_LOWERING=true` state working on my own and removing the flag once we're done alpha-testing JuliaLowering.

aviatesk:
Let's add the entry to make sure this flag keeps working as expected. I may enable it on my end too.

PR #189, opened 2025-07-20
Improve robustness to lowering failures

mlechu:
Some fixes from normal use:

- Even assuming perfect macro interop, there are a number of reasons macro expansion can fail (methoderror, incomplete input, analysis lag...), so if expansion fails, trim K"macrocall"s and try again.

- Trim K"error" before lowering (input still might be bad, but will be less likely to fail)

Also hide bindings with scope_layer != 1 (unescaped vars from a macro expansion) or with `#` in the name (either manual gensyms, or internals leaking out)


aviatesk (approved):
Good enhancements!

PR #194, opened 2025-07-24
formatting: Add Runic integration

aviatesk:


aviatesk:
I think this PR is in good condition.
Regarding the tests, I will add them after the configuration system (#185) is merged.

PR #201, opened 2025-08-05
diagnostics: report errors from JuliaLowering

aviatesk:
- [x] Report lowering errors
- [x] Report macro-not-found errors
- [x] Report macro expansion errors
- [x] Remove duplicated errors from JET's top-level analysis
- [x] Show inner macro expansion error

### DIFF
--- a/src/diagnostics.jl
+++ b/src/diagnostics.jl
@@ -185,11 +185,17 @@ function lowering_diagnostics!(diagnostics::Vector{Diagnostic}, st0::JL.SyntaxTr
         elseif err isa JL.MacroExpansionError
             st = scrub_expand_macro_stacktrace(stacktrace(catch_backtrace()))
             msg = err.msg
-            if msg == "Error evaluating macro name"
-                name = JS.sourcetext(err.ex)
-                msg = "Macro name `$name` not found"
-                relatedInformation = nothing
-            else
+            relatedInformation = missing
+            if isdefined(err, :err)
+                inner = err.err
+                if msg == "Error evaluating macro name" && inner isa UndefVarError
+                    msg = "Macro name `$(inner.var)` not found"
+                    relatedInformation = nothing
+                else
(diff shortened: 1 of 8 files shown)


src/diagnostics.jl
```diff
@@ -141,8 +141,18 @@ function lowering_diagnostics!(diagnostics::Vector{Diagnostic}, st0::JL.SyntaxTr
...
+                #=severity=#DiagnosticSeverity.Error,
+                #=source=#LOWERING_DIAGNOSTIC_SOURCE))
+        elseif err isa JL.MacroExpansionError
+            push!(diagnostics, jsobj_to_diagnostic(err.ex, sourcefile, err.msg,
+                #=severity=#DiagnosticSeverity.Error,
+                #=source=#LOWERING_DIAGNOSTIC_SOURCE))
+        else
+            JETLS_DEBUG_LOWERING && @warn "Error in lowering" err
+            JETLS_DEBUG_LOWERING && Base.show_backtrace(stderr, catch_backtrace())
+        end
```
aviatesk: @mlechu Is it possible to raise a dedicated exception from JL when a macro isn't found? Currently, an `UndefVarError` seems to be raised, but that makes it hard to distinguish it from a potential `UndefVarError` that might come from within JuliaLowering itself. So in cases like the following, I'd like to catch an error like `MacroNotFoundException` that holds a reference to the `SyntaxTree` of the relevant macro.
```julia
julia> st = jlparse("x = @notexisting 42");

julia> res = try
           JETLS.jl_lower_for_scope_resolution(st[1], Main);
       catch e
           e
       end
UndefVarError(Symbol("@notexisting"), 0x00000000000097d2, Main)
```

aviatesk: Implemented at: https://github.com/c42f/JuliaLowering.jl/pull/25.

PR #203, opened 2025-08-07
diagnostics: add guard for code generated by old macros

aviatesk:
The current `#jetls-hacking` branch can lower code using old macros, but the output `SyntaxTree` loses provenance information in those cases. `JS.first_byte` and `JS.last_byte` seem to return `0` always.

Added guards to handle these cases as a temporary measure.

@mlechu Is it expected that `JS.[first|last]_byte` return `0` in those cases?


mlechu (approved):

mlechu:
It is expected; see https://github.com/c42f/JuliaLowering.jl/blob/fedc5c5d4bfdafd2d6818892f47fe3963d1c02c9/src/syntax_graph.jl#L303-L316

PR #205, opened 2025-08-07
remove dependency on DataStructures.jl by managing sorted configs manually

abap34:
Removed usage of `DataStructures.SortedDict` and now maintain a sorted array of configs manually. This slightly worsens theoretical runtime performance, but the actual impact is negligible due to the small number of config files. This change removes the dependency on DataStructures.jl.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -223,7 +223,32 @@ struct WatchedConfigFiles
     configs::Vector{Dict{String,Any}}
 end
 
-WatchedConfigFiles() = WatchedConfigFiles(String[], Dict{String,Any}[])
+# TODO (later): move this definition to external files
+global DEFAULT_CONFIG::Dict{String,Any} = Dict{String,Any}(
+    "performance" => Dict{String,Any}(
+        "full_analysis" => Dict{String,Any}(
+            "debounce" => 1.0,
+            "throttle" => 5.0
+        )
+    ),
+    "testrunner" => Dict{String,Any}(
+        "executable" => "testrunner"
+    ),
+)
+
+global CONFIG_RELOAD_REQUIRED::Dict{String,Any} = Dict{String,Any}(
+    "performance" => Dict{String,Any}(
+        "full_analysis" => Dict{String,Any}(
+            "debounce" => true,
+            "throttle" => true
+        )
+    ),
+    "testrunner" => Dict{String,Any}(
+        "executable" => false
+    ),
+)
+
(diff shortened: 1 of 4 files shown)


test/utils/test_config.jl
```diff
@@ -47,6 +47,92 @@ const TEST_RELOAD_REQUIRED = Dict{String, Any}(
...
+        @test JETLS._file_idx(watched, "___UNDEFINED___") === nothing
+        watched["/project/.JETLSConfig.toml"] = config
+        idx = JETLS._file_idx(watched, "/project/.JETLSConfig.toml")
+        @test idx !== nothing
+        @test idx == 1
+        @test watched.files[idx] == "/project/.JETLSConfig.toml"
+        @test watched.configs[idx] == config
+    end
+end
+
```
aviatesk:
```suggestion
end
```

Not a strong opinion, but since this is an implementation detail of `WatchedFiles` and not used anywhere else, I don't think we need to have unit tests for this.

src/types.jl
```diff
@@ -218,12 +218,63 @@ struct LSPostProcessor
...
+    end
+    return idx
+end
+
+Base.keys(watched_files::WatchedFiles) = watched_files.files
+Base.values(watched_files::WatchedFiles) = watched_files.configs
+Base.length(watched_files::WatchedFiles) = length(watched_files.files)
+Base.haskey(watched_files::WatchedFiles, file::String) = _file_idx(watched_files, file) !== nothing
+
+function Base.delete!(watched_files::WatchedFiles, file::String)
```
aviatesk: Maybe better to add a guard against deletion of `__DEFAULT_CONFIG__`? Also, if we use this data structure specifically for configuration files, we may want to add `__DEFAULT_CONFIG__` within the constructor of `WatchedFiles` (and maybe rename `WatchedFiles` to `WatchedConfigFiles`?).

abap34: I've added a guard and also changed the name. (Now, deleting`__DEFAULT_CONFIG__` is just skipped. It would be better to raise an error instead?)

aviatesk: > It would be better to raise an error instead?

Yeah, raising an explicit error is my preference.

What do you think about registering `__DEFAULT_CONFIG__` within the `WatchedConfigFiles()` constructor so that we don't need to register it from the call site? If you think this complicates things, I'm just fine with taking this back.

abap34: > Yeah, raising an explicit error is my preference.

I've updated it. 663fe7d

> What do you think about registering `__DEFAULT_CONFIG__` within the `WatchedConfigFiles()` constructor 

I agree that it makes more sense. It required some rearrangement, but since the logic is mostly unchanged, it wasn't too complicated. I've made the change. acd1666

PR #206, opened 2025-08-08
use more backtick code fences for `showprov` output safety

abap34:
The `showprov` function used for hover displays includes several lines before and after the definition. If any of those lines contain a literal triple backtick (` ``` `), it can severely break the Markdown layout.

For example, code like the following is very common:

````julia
"""
    func(x)

description

```julia
julia> func(1)
2
```
"""
func(x) = 1 + x
````

However, hovering over the line `func(x) = 1 + x` will cause layout issues due to the triple backtick two lines above.

Fortunately, in most cases (at least in CommonMark and GitHub Flavored Markdown), code blocks allow any number of backticks ≥ 3. So by simply increasing the outer fence length, we can avoid these collisions in most cases. There may be stricter and more precise ways to fix this, but for now, I think this patch provides a simple and effective solution.

|              | VSCode | Helix |
|--------------|--------|-------|
| **Before**   | VSCode before | VSCode after |
| **After**    | helix before | helix after |


abap34:
I hadn't noticed that. Thank you!  I've added it now.

PR #228, opened 2025-08-22
use `null` instead of an empty array when no highlights are found

abap34:
Some clients raise an error if an empty array is returned when there are no highlights (confirmed in VSCode). It appears that returning `null` is the expected response.


aviatesk (approved):
The spec doesn't actually prevent us from returning an empty highlight column. So I think it's a bug on the VSCode side, but oh well.

PR #235, opened 2025-08-26
Update JuliaLowering branch to master (with SyntaxGraph changes)

mlechu:
JuliaLowering can run everything in compatibility mode at this point, so we may not even need `jetls-hacking` anymore.  As of writing, this is a version bump to https://github.com/c42f/JuliaLowering.jl/pull/35, which is the master branch plus some syntax graph utilities.  I've still created a separate jetls branch for stability and in case you'd like to apply any changes (@aviatesk)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 [sources]
 JET = {rev = "master", url = "https://github.com/aviatesk/JET.jl"}
 JSONRPC = {path = "JSONRPC"}
-JuliaLowering = {rev = "jetls-hacking", url = "https://github.com/mlechu/JuliaLowering.jl"}
+JuliaLowering = {rev = "jetls-2", url = "https://github.com/mlechu/JuliaLowering.jl"}
 JuliaSyntax = {rev = "jetls-hacking", url = "https://github.com/JuliaLang/JuliaSyntax.jl"}
 LSP = {path = "LSP"}
 
--- a/src/diagnostics.jl
+++ b/src/diagnostics.jl
@@ -229,7 +229,7 @@ function lowering_diagnostics!(
 
         st0 = without_kinds(st0, JS.KSet"error macrocall")
         try
-            ctx1, st1 = JL.expand_forms_1(mod, st0)
+            ctx1, st1 = JL.expand_forms_1(mod, st0, true)
             _jl_lower_for_scope_resolution(ctx1, st0, st1)
         catch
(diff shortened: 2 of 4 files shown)


mlechu:
The one thing this new branch is missing compared to `jetls-hacking` is the toplevel-expr-from-macro fix (https://github.com/mlechu/JuliaLowering.jl/commit/80ba20ad8fd1effb8ab4755e8dce0b2b1f5757d4).  This will be fixed on master eventually, but if it causes too much trouble in the meantime (it would probably present as `@doc` weirdness), we can apply this change too.

aviatesk (approved):
Thanks for the update. I basically want to keep up with JL's `#main`, so I totally agree with using a branch based on the latest `#main`.

> the toplevel-expr-from-macro fix ([mlechu/JuliaLowering.jl@80ba20a](https://github.com/mlechu/JuliaLowering.jl/commit/80ba20ad8fd1effb8ab4755e8dce0b2b1f5757d4))

I don't quite understand what cases this fix is for. Is it needed for the expansion of macro-defining-macrocalls, like what's being discussed in JuliaLang/julia#53515, or is it for a broader set of cases?

mlechu:
The original change https://github.com/JuliaLang/julia/pull/53515 is needed for correctness in macro-defining-macrocalls, yes.  I think there might be other reasons we need to save the current expansion module in the `(hygienic-scope ,expression ,module)` form for later expansion, but I can't think of an easy example, so I'm not sure.

The reason this is relevant outside of edge cases is because `:toplevel`-returning macros have now adapted to this change, so may contain extra `:escape`s that cause errors if we don't add the expected `hygienic-scope`s.  Easy to do what we've been doing for Expr, needs some thinking for SyntaxTree.

aviatesk:
Got it. So the most straightforward and perceivable example is probably that an expression like `@enum A a=1` can't be lowered. It seems the current `#jetls-hacking` can't do that either (though it might be a different error).

I'd always prefer to use `#main`, but it's totally possible that breaking changes could be made to JL's `#main` in the near future. For now, it might be better to keep using a custom branch, just like we've been doing. Would a new branch be better than just updating `#jetls-hacking`? Either way works for me.

Also please update `test/Project.toml` too. That's where the JL used by `Pkg.test()` is specified.

mlechu:
> Would a new branch be better than just updating #jetls-hacking? Either way works for me.

I needed a new branch to atomically update the lowering arguments here.  Feel free to force-overwrite `jetls-hacking` and switch to back that if you ever need to make a similar breaking change.

PR #246, opened 2025-09-10
fix docstring in `merge_config!`

abap34:
I thought I had submitted my review, but apparently I hadn’t. 😅
This just fixes a docstring that became outdated after a refactor (#242)  changed the function arguments.


aviatesk (approved):
Good catch!

PR #262, opened 2025-09-22
lifecycle: Add server self-shutdown mechanism when client process dies

aviatesk:
Monitors client processId every 60 seconds and sends a special self-shutdown token if client is no longer running, preventing orphaned server processes.

### DIFF
--- a/src/lifecycle.jl
+++ b/src/lifecycle.jl
@@ -218,6 +218,21 @@ function handle_InitializeRequest(server::Server, msg::InitializeRequest)
             name = "JETLS",
             version = "0.0.0"))
 
+    processId = init_params.processId
+    if !isnothing(processId)
+        Threads.@spawn while true
+            # To handle cases where the client crashes and cannot execute the normal
+            # server shutdown process, check every 60 seconds whether the `processId`
+            # is alive, and if not, put a special message token `SelfShutdownNotification`
+            # into the `endpoint` queue. See `runserver(server::Server)`.
+            sleep(60)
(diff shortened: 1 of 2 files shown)


topolarity:
Just want to express that I am *very* grateful for some robust lifecycle management 🙂 

It's one of my biggest gripes with the current LS that it can leave behind zombies that are very resource-hungry. I'm sure this doesn't 100% resolve the issue, but I'm glad you're working on it.

src/lifecycle.jl
```diff
@@ -218,6 +218,21 @@ function handle_InitializeRequest(server::Server, msg::InitializeRequest)
...
 
+    processId = init_params.processId
+    if !isnothing(processId)
+        Threads.@spawn while true
+            # To handle cases where the client crashes and cannot execute the normal
+            # server shutdown process, check every 60 seconds whether the `processId`
+            # is alive, and if not, put a special message token `SelfShutdownNotification`
+            # into the `endpoint` queue. See `runserver(server::Server)`.
+            sleep(60)
+            if !iszero(@ccall uv_kill(processId::Cint, 0::Cint)::Cint)
```
topolarity, line 229: There is a bit of a race condition here IIUC where the process might be killed and then the PID is re-allocated to another process, right?

aviatesk: Yes, that's right. I briefly looked into more accurate ways to check if a client process is alive or dead, but IIUC there is no complete answers for this, and they require platform-dependent code, and I wasn't confident I could test them well. So, for now, I've merged this simplest approach. I hope this should cover most cases (since processes might not always be reused, and even if they are, the new process might just die soon after).

PR #271, opened 2025-10-13
Emacs/eglot: use TCP socket for communication

rossviljoen:
As far as I can tell, eglot can't use Unix domain sockets, but it can use TCP.

Ref: https://www.gnu.org/software/emacs/manual/html_mono/eglot.html#index-eglot_002dserver_002dprograms

Feel free to close if this isn't a better default

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,9 @@ Minimal [Emacs](https://www.gnu.org/software/emacs/)
                 "--history-file=no"
                 "--project=/path/to/JETLS.jl"
                 "--threads=auto"
-                "/path/to/JETLS.jl/runserver.jl"))
+                "/path/to/JETLS.jl/runserver.jl"
+                "--socket"
+                :autoport))
 ```
 ### Neovim
 


README.md
```diff
@@ -130,7 +130,9 @@ Minimal [Emacs](https://www.gnu.org/software/emacs/)
                 "--history-file=no"
                 "--project=/path/to/JETLS.jl"
                 "--threads=auto"
-                "/path/to/JETLS.jl/runserver.jl"))
+                "/path/to/JETLS.jl/runserver.jl"
+                "--socket"
+                :autoport))
```
aviatesk, line 135: The `socket` argument is expected to be a parsable integer, but does the `eglot` client automatically rewrite its `:autoport` argument appropriately?

rossviljoen: It does, yes. eglot picks a free port and then passes the resulting integer as an argument to the server. I've tested this locally and it seems to work well

aviatesk: That's great. Thanks for suggesting a great solution.

PR #280, opened 2025-10-16
add Pkg.jl app

ExpandingMan:
This copies `runserver.jl` into a new module `CLI` which is exposed as an [app](https://pkgdocs.julialang.org/v1/apps/).

I realize that this might be incompatible with some existing workflow to compile with JuliaC, so I realize it may not be sensible to merge this.  I have kept `runserver.jl` so as not to completely break compatibility.

If you feel it makes sense to merge this, I could update the README to reflect it.  If not, you can close this and if you feel it's appropriate I can open an issue for this.

### DIFF
--- a/runserver.jl
+++ b/runserver.jl
@@ -1,0 +1,178 @@
+module var"##__JETLSEntryPoint__##"
+
+@info "Running JETLS with Julia version" VERSION
+
+using Pkg
+using Sockets
+
+# TODO load Revise only when `JETLS_DEV_MODE` is true
+try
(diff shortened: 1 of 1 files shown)


aviatesk:
Sorry for the late feedback. We decided to switch to launching JETLS using the Pkg app, which meant we also needed to update the VSCode client implementation and documentation. We've completed that work in this [PR](https://github.com/aviatesk/JETLS.jl/pull/315). Sorry I couldn't merge this, but thanks for submitting it anyway.

PR #282, opened 2025-10-22
config: use struct instead of Dict

abap34:
## Summary

This PR changes the configuration storage object from a `Dict` to a struct (`JETLSConfig`).  
This enables:
- Better compiler understanding of configuration fields, improving LS performance  
- Automatic schema generation for completion and validation

## Implementation Details

A new dependency, [Configurations.jl](https://github.com/Roger-luo/Configurations.jl), has been introduced.  
This package maps TOML-parsed `Dict`s to structs.  
It’s relatively lightweight, so adding it as a dependency should not be a significant issue.

Since Configurations.jl does not currently provide full key paths for nested `KeyError`s,  
the helper function `collect_unmatched_key` from the previous implementation remains temporarily for error reporting.  
I plan to submit a PR to Configurations.jl to support this feature, so that, this workaround can be removed.

## Schema Generation

With [JSONSchemaGenerator.jl](https://github.com/matthijscox/JSONSchemaGenerator.jl),  
a JSON Schema can be generated simply as follows:

```julia
StructTypes.omitempties(::Type{T}) where {T <: ConfigSection} = fieldnames(T)
JSON.json(JSONSchemaGenerator.schema(JETLSConfig), 4)
```

Using this approach, completion and type checking become available in various editors.  
(The example below uses VSCode with [even-better-toml](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml).)

config_schema_demo_0

In the long term, it would be convenient to have the schema generated in CI at release time and registered to [schemastore.org](https://schemastore.org).  This part is not critical for the LS itself, so we can be more liberal in using convenient tools and dependencies to improve the workflow.

I would be happy to discuss this further later on :)

### DIFF
--- a/test/test_config.jl
+++ b/test/test_config.jl
@@ -218,6 +218,14 @@ end
     @test JETLS.get_config(manager, :testrunner, :executable) === "test_runner"
     @test JETLS.get_config(manager, :non_existent_key) === nothing
 
+    @test Base.infer_return_type((typeof(manager),)) do manager
+           JETLS.get_config(manager, :internal, :dynamic_setting)
+    end == Union{Nothing, Int}
+
+    @test Base.infer_return_type((typeof(manager),)) do manager
+           JETLS.get_config(manager, :internal, :static_setting)
+    end == Union{Nothing, Int}
+
     # Test priority: __DEFAULT_CONFIG__ has lower priority
     override_config = JETLS.JETLSConfig(;
         full_analysis=JETLS.FullAnalysisConfig(999.0),


src/config.jl
```diff
@@ -211,19 +173,20 @@ Retrieves the current configuration value.
 Among the registered configuration files, fetches the value in order of priority (see `Base.lt(::ConfigFileOrder, path1, path2)`).
 If the key path does not exist in any of the configurations, returns `nothing`.
 """
-function get_config(manager::ConfigManager, key_path::String...)
-    is_static_setting(key_path...) &&
-        return access_nested_dict(load(manager).static_settings, key_path...)
-    for config in values(load(manager).watched_files)
-        return @something access_nested_dict(config, key_path...) continue
+function get_config(manager::ConfigManager, key_path::Symbol...)
```
abap34: Should we use macros or something that expand to direct field access to make type inference results more precise?

aviatesk: For maximum type stability:
```diff
diff --git a/src/config.jl b/src/config.jl
index be50714..a1ed28c 100644
--- a/src/config.jl
+++ b/src/config.jl
@@ -173,7 +173,7 @@ Retrieves the current configuration value.
 Among the registered configuration files, fetches the value in order of priority (see `Base.lt(::ConfigFileOrder, path1, path2)`).
 If the key path does not exist in any of the configurations, returns `nothing`.
 """
-function get_config(manager::ConfigManager, key_path::Symbol...)
+Base.@constprop :aggressive function get_config(manager::ConfigManager, key_path::Symbol...)
     try
         is_static_setting(key_path...) &&
             return getobjpath(load(manager).static_settings, key_path...)
diff --git a/src/utils/general.jl b/src/utils/general.jl
index 404c5ff..336b59b 100644
--- a/src/utils/general.jl
+++ b/src/utils/general.jl
@@ -118,7 +118,7 @@ may be unset (set to `nothing`) depending on client/server capabilities.
 Traverses the field chain `paths...` of `obj`, and returns `nothing` if any
 `nothing` field is encountered along the way.
 """
-function getobjpath(obj, path::Symbol, paths::Symbol...)
+@noinline Base.@constprop :aggressive function getobjpath(obj, path::Symbol, paths::Symbol...)
     nextobj = @something getfield(obj, path) return nothing
     getobjpath(nextobj, paths...)
 end
```

src/utils/general.jl
```diff
@@ -117,8 +117,11 @@ An accessor primarily written for accessing fields of LSP objects whose fields
 may be unset (set to `nothing`) depending on client/server capabilities.
 Traverses the field chain `paths...` of `obj`, and returns `nothing` if any
 `nothing` field is encountered along the way.
+
+Note: `@noinline` is used to maximize type stability. This is a temporary workaround and
+may become unnecessary with future compiler improvements.
```
aviatesk, line 122: It might be best to add test case like
```julia
julia> @test Base.infer_return_type((typeof(manager),)) do manager
           JETLS.get_config(manager, :full_analysis, :debounce)
       end == Union{Nothing,Float64}
```

abap34: resolved in b7aa529

src/config.jl
```diff
@@ -106,49 +17,96 @@ function Base.lt(::ConfigFileOrder, path1, path2)
...
-    for (k, v) in dict
-        if v isa ConfigDict
-            cleaned_v = cleanup_empty_dicts(v)
-            if isempty(cleaned_v)
-                result = Base.delete(result, k)
-            elseif cleaned_v != v
-                result = ConfigDict(result, k => cleaned_v)
-            end
-        end
+@generated function on_difference(
```
aviatesk, line 20: Maybe we can do something like:
```julia
using InteractiveUtils: InteractiveUtils
for T in InteractiveUtils.subtypes(ConfigSection)
    @eval function on_difference(
        callback,
        old_config::$T,
        new_config::$T,
        path::NTuple{N,Symbol}=()
    ) where {N}
        $T($((
            :(on_difference(
                callback,
                getfield(old_config, $(QuoteNode(fname))),
                getfield(new_config, $(QuoteNode(fname))),
                (path..., $(QuoteNode(fname)))
            ))
            for fname in fieldnames(T)
        )...))
    end

    @eval function on_difference(
        callback,
        old_val::$T,
        new_val::Nothing,
        path::Tuple
    )
        $T($((
            :(on_difference(
                callback,
                getfield(old_val, $(QuoteNode(fname))),
                nothing,
                (path..., $(QuoteNode(fname)))
            ))
            for fname in fieldnames(T)
        )...))
    end

    @eval function on_difference(
        callback,
        old_val::Nothing,
        new_val::$T,
        path::Tuple
    )
        $T($((
            :(on_difference(
                callback,
                nothing,
                getfield(new_val, $(QuoteNode(fname))),
                (path..., $(QuoteNode(fname)))
            ))
            for fname in fieldnames(T)
        )...))
    end
end
```
?

aviatesk: `@generated` might be cleaner though.

aviatesk: I think `@generated` might be just better since it works nicer than this with Revise.

aviatesk:
Thanks so much for working on this, @abap34

PR #284, opened 2025-10-23
formatting: add support for JuliaFormatter

aviatesk:
This commit extends the formatting system to support multiple formatter backends beyond Runic.jl. Users can now choose between preset formatters (`"runic"` or `"jlfmt"`) or configure custom formatters with their own executables.

The formatter configuration has been redesigned from a nested structure (`formatter.runic.executable`) to a simpler top-level `formatter` setting that accepts either:
- A preset name string: `"runic"` or `"jlfmt"`
- A custom formatter defined under `formatter.custom` with `executable` and optional `executable_range` fields

Due to limitations in Configurations.jl, custom formatters must be specified in the `[formatter.custom]` section rather than directly as the `formatter` value. Custom formatters can specify separate executables for document and range formatting operations.

Additionally, `FormattingOptions` from LSP requests are now propagated through the formatting pipeline, enabling formatters to respect client-provided settings like tab size. For JuliaFormatter, the tab size option is passed via the `--indent` flag. This client-wise setting will be overwritten by .JuliaFormatter.toml configuration if it's found since we are using the `--prioritize-config-file` option.

Requires https://github.com/domluna/JuliaFormatter.jl/pull/947 to be merged.

### DIFF
--- a/src/formatting.jl
+++ b/src/formatting.jl
@@ -3,9 +3,11 @@ const FORMATTING_REGISTRATION_METHOD = "textDocument/formatting"
 const RANGE_FORMATTING_REGISTRATION_ID = "jetls-rangeFormatting"
 const RANGE_FORMATTING_REGISTRATION_METHOD = "textDocument/rangeFormatting"
 const RUNIC_INSTALLATION_URL = "https://github.com/fredrikekre/Runic.jl#installation"
+const JULIAFORMATTER_INSTALLATION_URL = "https://github.com/domluna/JuliaFormatter.jl#installation"
 
 struct FormattingProgressCaller <: RequestCaller
     uri::URI
+    options::FormattingOptions
     msg_id::Union{String,Int}
     token::ProgressToken
 end
@@ -13,6 +15,7 @@ end
 struct RangeFormattingProgressCaller <: RequestCaller
     uri::URI
     range::Range
+    options::FormattingOptions
     msg_id::Union{String,Int}
     token::ProgressToken
 end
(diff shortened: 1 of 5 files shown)


src/formatting.jl
```diff
@@ -103,48 +108,109 @@ function do_format_with_progress(
...
 end
 
-function format_result(state::ServerState, uri::URI)
+function get_formatter_executable(formatter::FormatterConfig, for_range::Bool)
+    if formatter isa String
+        # Preset formatter: "runic" or "jlfmt"
+        if for_range
+            executable = default_executable(formatter)
+            if executable === nothing
+                if formatter == "jlfmt"
```
abap34: `executable` becomes `nothing` only when the formatter is neither `"runic"` nor `"jlfmt"`, so this branch looks unreachable.
It would be cleaner to first check whether a common preset exists regardless of `for_range`, and then perform the `"jlfmt"` check like this?:

```julia
function get_formatter_executable(formatter::FormatterConfig, for_range::Bool)
    if formatter isa String
        executable = default_executable(formatter)
        executable === nothing &&
            return request_failed_error("Unknown formatter preset...")
        if for_range && formatter == "jlfmt"
            return request_failed_error(
                "JuliaFormatter does not support range formatting...")
        else
            ...

File: src/types.jl [hunk 2]
```diff
@@ -334,18 +334,36 @@ end
...
+    if formatter == "runic"
+        return @static Sys.iswindows() ? "runic.bat" : "runic"
+    elseif formatter == "jlfmt"
+        return @static Sys.iswindows() ? "jlfmt.bat" : "jlfmt"
+    else
+        return nothing
+    end
+end
+
+function default_executable_range(formatter::String)
```
abap34: It looks unused function

aviatesk: Good catch.

src/formatting.jl
aviatesk: Yeah, that sounds great.

PR #295, opened 2025-10-29
Small fixes for CI

mlechu:
Mark one test as broken, and add an argument to a JL function

### DIFF
--- a/test/test_diagnostics.jl
+++ b/test/test_diagnostics.jl
@@ -153,7 +153,7 @@ end
                     break
                 end
             end
-            @test_broken found_diagnostic
+            @test found_diagnostic
         end
     end
 end


test/test_diagnostics.jl
```diff
@@ -153,7 +153,7 @@ end
                     break
                 end
             end
-            @test found_diagnostic
+            @test_broken found_diagnostic
```
aviatesk: Not sure why this test case is broken now? Looks like this doesn't relate to JL.

mlechu: It does look unrelated, probably a JET change? 

I marked it as broken because I thought it had been broken for longer than it was.  Feel free to unmark it.

aviatesk: Looks like I broke this test case recently. Already fixed up on master.

aviatesk:
```suggestion
            @test found_diagnostic
```

aviatesk (approved):

PR #297, opened 2025-11-13
comm: migrate to JSON.jl v1

aviatesk:


LSP/src/base-protocol.jl
```diff
@@ -3,16 +3,29 @@
...
+        ancestor_stack::Union{Nothing, Vector{Any}}, io::Union{Nothing, IO},
+        ind::Int, depth::Int, bufsize::Int
+    )
+    @assert length(buf) >= pos + 3
+    buf[pos+0] = UInt8('n')
+    buf[pos+1] = UInt8('u')
+    buf[pos+2] = UInt8('l')
+    buf[pos+3] = UInt8('l')
+    return pos + 4
+end
```
aviatesk: This is currently a hack because JSON.jl doesn't let us specify nullable fields per-field. But good news, we expect to remove this thanks to https://github.com/JuliaIO/JSON.jl/pull/408.

PR #306, opened 2025-11-19
use byte offset for UTF-8 position encoding instead of character count 

abap34:
This PR fixes issue when communicating with editors that expect UTF-8–based offsets. The previous implementation counted characters, but according to the LSP specification, offsets in UTF-8 mode should be counted in bytes:
[https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=Character%20offsets%20count%20UTF%2D8%20code%20units%20(e.g%20bytes)](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=Character%20offsets%20count%20UTF%2D8%20code%20units%20%28e.g%20bytes%29)

This PR updates the offset calculation accordingly and fixes the misalignment observed in editors such as Helix. (see below)
Editors that use UTF-16 by default (e.g., VS Code) are not affected.

| before  | after |
| ---- | ---- |
| image | image |

### DIFF
--- a/src/utils/string.jl
+++ b/src/utils/string.jl
@@ -34,7 +34,7 @@ function pos_to_utf8_offset(s::String, ch::UInt, encoding::PositionEncodingKind.
             offset = nextind(s, offset)
         end
     elseif encoding == PositionEncodingKind.UTF8 # UTF-8 counts bytes
-        offset = Int(ch) + 1
+        offset = min(Int(ch), sizeof(s)) + 1
     else # UTF-32 counts characters
         while offset <= sizeof(s) && char_count < ch
             char_count += 1
--- a/test/utils/test_string.jl
+++ b/test/utils/test_string.jl
@@ -59,15 +59,21 @@ using JETLS.LSP.PositionEncodingKind: UTF8, UTF16, UTF32
         @test pos_to_utf8_offset("", UInt(0), UTF16) == 1
         @test pos_to_utf8_offset("", UInt(10), UTF16) == 1
         @test pos_to_utf8_offset("", UInt(0), UTF8) == 1
+        @test pos_to_utf8_offset("", UInt(10), UTF8) == 1  # Clamped to end
 
-        # Position beyond string
+        # Position beyond string - UTF-8 should clamp to string bounds
         @test pos_to_utf8_offset("ab", UInt(10), UTF16) == 3
-        @test pos_to_utf8_offset("ab", UInt(10), UTF8) == 11  # Byte 10
+        @test pos_to_utf8_offset("ab", UInt(10), UTF8) == 3  # Clamped to sizeof("ab")+1
+        @test pos_to_utf8_offset("abc", UInt(100), UTF8) == 4  # Clamped to end
+
+        # Position exactly at end of string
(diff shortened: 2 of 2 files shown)


aviatesk:
Thanks for the fix! When I implemented aviatesk/JETLS.jl#207, I thought UTF-8 could be treated the same way as UTF-32, but it seems that's not the case (which is obvious now that you mention it).

src/utils/string.jl
```diff
@@ -15,9 +15,10 @@ Convert a character position to a UTF-8 byte offset in a Julia string.
 1-based byte offset in the UTF-8 encoded string `s`
 
 # Details
-- For UTF-8: `ch` represents character count (same as `length(s)`)
+According to the LSP specification:
+- For UTF-8: `ch` represents byte offset (0-based)
```
aviatesk: This "0-based" comment doesn't seem to be a condition specific to the `UTF-8` case, so it appears a bit random to me.

src/utils/string.jl
```diff
@@ -32,7 +33,9 @@ function pos_to_utf8_offset(s::String, ch::UInt, encoding::PositionEncodingKind.
             end
             offset = nextind(s, offset)
         end
-    else # UTF-8 or UTF-32 both count characters
+    elseif encoding == PositionEncodingKind.UTF8 # UTF-8 counts bytes
+        offset = Int(ch) + 1
```
aviatesk: Do we need to handle the case of exceeding the end of string here?
I think there are guards implemented for the other encodings to prevent exceeding `sizeof(s)` (having said that I've honestly forgotten why that was necessary though).

src/utils/string.jl
abap34: That’s indeed true. 8261204

src/utils/string.jl
abap34: I also thought it might be better to clamp it, for example with something like:
`offset = min(Int(ch) + 1, sizeof(s) + 1)`

As far as I remember, I once investigated this because I was worried about potential mismatches between characters and positions in cases like:

```
\pr    # user triggers completion here
```

and then the user keeps typing:

```
\prim  # user finishes typing here then accepts a completion
```

In practice, the client handled this correctly, so it wasn’t an issue. However, having to investigate client behavior for cases like this is something I’d prefer to avoid, and I believe you would agree. It seems reasonable for the server to take responsibility for clamping positions safely.

aviatesk: Yeah, most client probably has similar guards, but it wouldn't hurt for us to add them on our side too. It's already implemented in other code paths, so let's add the guard.

src/utils/string.jl
```diff
@@ -96,8 +99,8 @@ Convert a 1-based byte offset to a 0-based line and character number.
 
 # Returns
 `pos::Position` where both are 0-based indices, with character position calculated according
-to the specified encoding:
-- UTF-8: Character count (each Unicode character = 1 unit)
+to the specified encoding per LSP specification:
+- UTF-8: Byte offset within the line (0-based)
```
aviatesk:
```suggestion
- UTF-8: Byte offset within the line
```

src/utils/string.jl
abap34: 07892eb

Issue #358, opened 2025-12-05
Code lens response is empty until file analysis has completed


jwortmann:
Similar to #333, just for code lenses.

When a file containing tests is opened for the first time after starting the server, the response for a `textDocument/codeLens` request is empty. If another request is sent by the client (e.g. when the file content is modified or after closing and re-opening the file) after the file analysis has completed, code lenses are reported correctly. This creates a confusing UX for users when code lenses seem to work only "sometimes".

I can think of 2 possible solutions:

1. Defer calculating the code lenses and don't send the response until the file analysis has completed. There is nothing wrong with keeping a request open until correct results are available. If the file content changed in the meanwhile, the client will likely cancel the open request and send a new one after `didChange`.

2. Send a [`workspace/codeLens/refresh`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_refresh) request after the file analysis has completed. The problem with this approach is that it is not scoped to a particular file, and therefore the client has to send new requests for all opened (or visible) files. See the description from the specs that explains that the refresh requests are intended to be used on project configuration changes:
    ```typescript
    /**
	 * Note that this event is global and will force the client to refresh all
	 * code lenses currently shown. It should be used with absolute care and is
	 * useful for situation where a server for example detect a project wide
	 * change that requires such a calculation.
	 */
    ```

(2b.) Like 2., but only send the refresh request if you have detected that there are code lenses in the file.

aviatesk:
Thank you for raising this issue.
I decided to resolve this problem by implementing your proposal 1 in #362. Generalizing the solution to this problem became the catalyst for fixing many other issues.

jwortmann:
Thanks for working on it. However, using the latest update `JETLS_VERSION = "2025-12-05"` I can still see the same behavior as before, except for the very first file that was opened when initializing the server.

For instance, I first open some file from the `src` directory of a package and wait until the server fully initializes and stops sending progress notifications. Then I open the file `test/runtests.jl` where I expect code lenses to be present, but the server immediately sends a response `null` for the `textDocument/codeLens` request (request ID 8 in the logs below). I need to edit or re-open the file to trigger a new code lens request in order to receive the correct response.

Relevant server logs:
```
[...]
[...]

Issue #327, opened 2025-11-25
ERROR: KeyError: key


ufechner7:
If I checkout the package KiteUtils.jl
```
git clone https://github.com/OpenSourceAWE/KiteUtils.jl
cd KiteUtils.jl
julia -e 'using Pkg; Pkg.update()'
```
and then open the folder KiteUtils.jl in VSCode, and then open the file `src/KiteUtils.jl` and then wait, I get the following error:
```
[jetls-client] Auto-selected communication channel: pipe
[jetls-client] Using communication channel: pipe
[jetls-client] Server listening on Unix domain socket: /tmp/jetls-121295-1764096019044.sock
[jetls-client] Starting JETLS...
[jetls-client] JETLS connected!
[JETLS-stderr] ┌ Info: Connected to Unix domain socket
[JETLS-stderr] └   pipe_connect_path = "/tmp/jetls-121295-1764096019044.sock"
[JETLS-stderr] ┌ Info: Running JETLS with the following setup:
[JETLS-stderr] │   Sys.BINDIR = "/home/ufechner/.julia/juliaup/julia-1.12.2+0.x64.linux.gnu/bin"
[JETLS-stderr] │   pkgdir(JETLS) = "/home/ufechner/.julia/packages/JETLS/M2xQU"
[JETLS-stderr] │   Threads.nthreads() = 16
[JETLS-stderr] │   JETLS_DEV_MODE = false
[JETLS-stderr] │   JETLS_TEST_MODE = false
[JETLS-stderr] └   JETLS_DEBUG_LOWERING = false
[jetls-client] JETLS is ready!
[JETLS-stderr] ┌ Error: Error in `execute_analysis_request` for 
[JETLS-stderr] │   request = JETLS.AnalysisRequest(JETLS.PackageSourceAnalysisEntry("/home/ufechner/repos/KiteUtils.jl/Project.toml", file:///home/ufechner/repos/KiteUtils.jl/src/KiteUtils.jl, Base.PkgId(Base.UUID("90980105-b163-44e5-ba9f-8b1c83bb0533"), "KiteUtils")), file:///home/ufechner/repos/KiteUtils.jl/src/KiteUtils.jl, 0, JETLS.CancellableToken("##WorkDoneProgressCreateRequest_request_analysis_on_open!#297", JETLS.CancelFlag(false)), true, nothing, Base.Event(Base.GenericCondition(ReentrantLock()), false, false))
[...]

Issue #274, opened 2025-10-14
Various bugs


jakobnissen:
Again, please let me know if JETLS is too early in development for these bugreports to be usable.

This happens when opening `https://github.com/BioJulia/BGZFLib.jl` at commit e00e5d92 using JETLS from today.

#### Cannot identify package environment
Error:
```
[JETLS-stderr] ┌ Warning: Failed to identify package environment
[JETLS-stderr] │   pkgname = "BGZFLib"
[JETLS-stderr] └ @ JETLS ~/code/JETLS.jl/src/analysis/
```
Package has only registered dependencies, and is fully resolved and updated.
This occurs intermittently. All the other errors below also occur when this error does not happen.

#### False positive: Invalid function name
Error:
```
Invalid function nameJETLS - lowering
```
This is marked on the line `Base.:(<)(x::VirtualOffset, y::VirtualOffset) = getfield(x, :x) < getfield(y, :x)`

#### False positive: `Unused argument `io`JETLS - lowering`
In this function where `io` is clearly used:
```
function get_writer_sink_room(io::AbstractBufWriter)::MutableMemoryView{UInt8}
    buffer = get_buffer(io)
    while length(buffer) < MAX_BLOCK_SIZE
        n = grow_buffer(io)
        iszero(n) && throw(BGZFError(nothing, BGZFErrors.insufficient_writer_space))
        buffer = get_buffer(io)
    end
    return buffer
end
```
There are dozens of this false positive all over the code.

#### Highlighting does not track code when moved
If I add newlines, the highlighted areas do not move correspondingly. For example, when adding a newline abive the `get_writer_sink_room` function, it looks like this. Note the wrong placement of the blue underlines. The blue underline under `while ` is marked `Unused local binding `buffer`JETLS - lowering`.

<img width="425" height="257" alt="Image" src="https://github.com/user-attachments/assets/63833368-13f0-4c39-8dc6-89f855fd1932" />

#### Cache files not found
When opening a new file in the project, JETLS crashes with:
```
[Error - 13:18:59] Request textDocument/diagnostic failed.
  Message: File cache for file:///home/jakni/code/BGZFLib.jl/src/index.jl is not found
  Code: -32803 
[object Object]
[Error - 13:18:59] Document pull failed for text document file:///home/jakni/code/BGZFLib.jl/src/index.jl
  Message: File cache for file:///home/jakni/code/BGZFLib.jl/src/index.jl is not found
  Code: -32803 
[object Object]
```

[...]

Issue #273, opened 2025-10-13
Crash in vscode extension


langestefan:
Hi! Thanks for making it easier to test now that the vscode extension is available.

I was using it today for a couple of hours just programming like a human. Then I got bored, so I fired up claude and JETLS started barking at me.

This is the vscode output of it crashing:

<details Crash log>
  <summary><i>Crash log</i></summary>

```julia
[JETLS-stderr]   1 dependency successfully precompiled in 2 seconds. 27 already precompiled.
[JETLS-stderr] WARNING: Imported binding Positioning.calculate_deltat was undeclared at import time during import to ##JETVirtualModule#344.
[JETLS-stderr] WARNING: Imported binding Positioning.calculate_deltat was undeclared at import time during import to ##JETVirtualModule#353.
[JETLS-stderr] WARNING: Imported binding Positioning.calculate_deltat was undeclared at import time during import to ##JETVirtualModule#359.
[Error - 00:35:04] Request textDocument/diagnostic failed.
  Message: File cache for file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/Positioning.jl is not found
  Code: -32803 
[object Object]
[Error - 00:35:04] Document pull failed for text document file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/Positioning.jl
  Message: File cache for file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/Positioning.jl is not found
  Code: -32803 
[object Object]
[Error - 00:35:28] Request textDocument/diagnostic failed.
  Message: File cache for file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/noaa.jl is not found
  Code: -32803 
[object Object]
[Error - 00:35:28] Document pull failed for text document file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/noaa.jl
  Message: File cache for file:///home/langestefan/dev/solar/SolarPosition.jl/src/Positioning/noaa.jl is not found
  Code: -32803 
[object Object]
[Error - 00:35:55] Request textDocument/diagnostic failed.
[...]

PR #362, opened 2025-12-05
handlers: Wait for file cache population in request handlers

aviatesk:
Request handlers now use a 3-argument version of `get_file_info` that waits for the file cache to be populated instead of immediately returning an error. This fixes "file cache not found" errors that occurred when LSP requests arrived before the cache was ready, particularly right after opening files.

The new `get_file_info(state, uri, cancel_flag)` polls until the cache is available, with support for cancellation and a configurable timeout (default 120 seconds to account for JIT compilation on first requests).

- fixes aviatesk/JETLS.jl#273
- fixes aviatesk/JETLS.jl#274
- fixes aviatesk/JETLS.jl#327
- fixes aviatesk/JETLS.jl#358

### DIFF
--- a/src/utils/server.jl
+++ b/src/utils/server.jl
@@ -167,7 +167,7 @@ get_file_info(s::ServerState, t::TextDocumentIdentifier) = get_file_info(s, t.ur
 """
     get_file_info(
         s::ServerState, uri::URI, cancel_flag::CancelFlag;
-        timeout = 120.0, cancelled_error_data = nothing, cache_error_data = nothing
+        timeout = 30., cancelled_error_data = nothing, cache_error_data = nothing
     ) -> Union{FileInfo,ResponseError}
 
 Wait for cached `FileInfo` to become available, with cancellation and timeout support.
@@ -183,7 +183,7 @@ Returns a `ResponseError` in two cases:
 """
 function get_file_info(
         s::ServerState, uri::URI, cancel_flag::CancelFlag;
-        timeout::Float64 = 120.0, cancelled_error_data = nothing, cache_error_data = nothing
+        timeout::Float64 = 30., cancelled_error_data = nothing, cache_error_data = nothing
     )
     start = time()
     while true
@@ -192,7 +192,6 @@ function get_file_info(
         cache = get(load(s.file_cache), uri, nothing)
         cache !== nothing && return cache
         if time() - start > timeout
-            JETLS_DEV_MODE && @warn "Timeout waiting for file cache" uri
             return file_cache_error(uri;
                 data = cache_error_data)
         end
--- a/test/setup.jl
+++ b/test/setup.jl
@@ -75,7 +75,7 @@ function withserver(f;
     - `raw_res`: The raw response(s) sent by the server (or `nothing` if `read=0`)
     - `json_res`: The JSON-parsed response(s) from the server (or `nothing` if `read=0`)
     """
-    function writereadmsg(@nospecialize(msg); read::Int=1)
+    function writereadmsg(@nospecialize(msg); read::Int=1, check::Bool=true)
         @assert read ≥ 0 "`read::Int` must not be negative"
         LSP.writelsp(in, msg)
         raw_msg = take_with_timeout!(received_queue)
@@ -92,7 +92,7 @@ function withserver(f;
                 push!(json_res, LSP.readlsp(out))
             end
         end
-        @test isempty(received_queue) && isempty(sent_queue)
+        check && @test isempty(received_queue) && isempty(sent_queue)
         return (; raw_msg, raw_res, json_res)
     end
 
--- a/test/test_diagnostics.jl
+++ b/test/test_diagnostics.jl
@@ -171,9 +171,8 @@ end
     end
 end
 
-@testset "file cache error handling" begin
-    # Test requesting diagnostics for a file that hasn't been opened
-    # (i.e., no file cache exists)
+@testset "File cache error handling" begin
+    # Test requesting diagnostics for a file whose cache has not been populated yet
     withscript("# some code") do script_path
         uri = filepath2uri(script_path)
         withserver() do (; writereadmsg, id_counter)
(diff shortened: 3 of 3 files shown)


clason:
This removes the error from https://github.com/aviatesk/JETLS.jl/issues/274#issuecomment-3591899676, but doesn't trigger diagnostics (i.e., reloading the file removes all already reported diagnostics from the buffer).

aviatesk:
What kind of problem is that?
If you can provide a specific reproducer, may take a look into it.

clason:
What do you mean by "what kind of problem"? The reproducer is the same that used to trigger the file cache error (open a project file in Neovim, wait for diagnosis to be complete and errors reported, reload the file with `:e<cr>`).

aviatesk:
I meant what diagnostics were missing. I'm guessing it's probably the `lowering/*` ones, and if so, the problem is probably on the client side (neovim).

clason:
All of them, even an obvious typo in a variable. (I'm also not sure how you can so quickly punt this to the client, if things are reported correctly on master.)

clason:
But I can confirm that this removes the error on reloading a file, which is already a strict improvement over master. 

The fact that errors are not correctly reported (e.g., `inference/undef-local-var: function_name(signature...)` instead of the previous "human readable" error message on `master`) is another matter.

aviatesk:
When I cycle through closing and opening files in VSCode or Zed, the diagnostics remain intact (rather re-issued internally), so I think it's likely due to some interaction with that Neovim command.
The fact that diagnostics disappear only in some clients typically means that the client is not sending requests.

clason:
> The fact that diagnostics disappear only in some clients typically means that the client is not sending requests.

Or that a server is not implementing the specification correctly but only looking at what works in VS Code. In my experience as a Neovim maintainer, this is the much more frequent case (especially if it's a regression introduced in a PR). I can assure you that reloading the file triggers resending diagnostics with other servers.

But this is not the regression I am talking about here, since this also happens on master; the regression is _what_ diagnostics are sent (reloading or not).

Of course, I'm not saying that there isn't a bug, just that one shouldn't jump to conclusions too quickly. I'm happy to help you if you're interested in supporting Neovim as first-class client; if this is meant to be VS Code (or Zed) only, let me know and I won't bother you anymore.

aviatesk:
> Or that a server is not implementing the specification correctly but only looking at what works in VS Code. In my experience as a Neovim maintainer, this is the much more frequent case (especially if it's a regression introduced in a PR). I can assure you that reloading the file triggers resending diagnostics with other servers.

Of course, that possibility exists as well. On the other hand, I wouldn't be too surprised if workflow/functionality that appear to work with other servers doesn't work with JETLS.
Since JETLS actually uses two types of diagnostics channels (`textDocument/diagnostic`/`textDocument/publishDiagnostic`) for various reasons, depending on the type of diagnostics you have, your issue might occur when one of them is not re-issued during that reload cycle. However, other servers might be using either of the diagnostic channels, which happens to be the one supposed as first-class by Neovim, or perhaps other channels (`workspace/diagnostics`).

My conclusion may have sounded hasty, but it was just a conditional guess, i.e. if diagnostics for `lowering/*.` code are not being sent, then `textDocument/diagnostic` request is not being sent from the client, and in that case, JETLS is just responding (almost statelessly) to requests from the client, so the problem is largely on the Neovim side.

Looking closely at your comment, it seems to be `inference/undef-local-var`, which means that `textDocument/publishDiagnostic` is not being re-issued. But even in that case other clients display already published diagnostics for newly opened files as well, this seems like a problem that should be resolved on the client side (though this may be understandable since `publishDiagnostic` is sort of unrecommended in LSP now).

> I'm happy to help you if you're interested in supporting Neovim as first-class client; if this is meant to be VS Code (or Zed) only, let me know and I won't bother you anymore.

Of course, it's great that there are users seriously using JETLS with Neovim. I would like to keep JETLS as client-agnostic an implementation as possible, and your help has been useful.

clason:
I apologize for my harsh tone; I do appreciate all the hard work you put into this project, and I understand that everything is still under heavy development so things are moving fast and can easily regress (and that's a completely fine strategy)! I'm genuinely trying to help with making this server ready to be used, especially in Neovim (since I want to use it).

> Looking closely at your comment, it seems to be inference/undef-local-var, which means that textDocument/publishDiagnostic is not being re-issued. But even in that case other clients display already published diagnostics for newly opened files as well, this seems like a problem that should be resolved on the client side.

Yes, that is the problem with (supporting only) `publishDiagnostic` -- since it's purely controlled server-side, reloading the file (which invalidates _all_ cached buffer state -- that's the point of a reload!) in the client will be invisible. Not sure if it is possible (or reasonable) to add a special-case workaround here. In any case, if it's documented, it is what it is and users will have to live with it. Now that publishing diagnostics are correctly deferred until analysis is complete, this is no longer a necessary workaround.

~~But note that the point of my last comment was actually that the _content_ of the diagnostic messages has changed, which _is_ a significant regression from `master` that is unrelated to reloading (since it also happens on "normal" opening of a file).~~ This is fixed in the merged PR; apparently simply checking out the branch is not enough to test a PR, I'll have to look into this.

aviatesk:
> my harsh tone

No worries, my tone might have been a bit strong too, but that often happens in online conversations.

> Not sure if it is possible (or reasonable) to add a special-case workaround here.

Right, I didn't check what LSP messages are sent when Neovim reloads, but JETLS requests a new analysis for the parsing unit containing a file every time it receives a file open notification, which I guess what happens on reloading, but if the file hasn't changed, the analysis is simply cancelled. Even then, the previous `textDocument/publishDiagnostic` isn't cleared, so the client should still see the old diagnostics.

> apparently simply checking out the branch is not enough to test a PR

This might be because the environment used when JETLS is installed via the `#release` channel is quite different from a local development setup. If you check out for local development, you need to run `Pkg.resolve(); Pkg.install()` in the root directory.

clason:
Yeah, I think I forgot the `]update` step I usually do.

But it turns out this _hasn't_ gone away; it's just not consistent. So depending on which file I open, for the same hint (which is spurious, by the way, but that's another issue) I either get the "unused" message, the `kwcall` internal name, or an empty message.

aviatesk:
Yeah, that's a separate issue rooted from JET internals. We should fix it anyway.

clason:
(I'm sorry I can't provide a minimal example; I tried before but it seems to need some of the scaffolding of the full project.)

aviatesk:
No worries. I do encounter these issue on my end but simply couldn't have time to work on it.

PR #364, opened 2025-12-05
Bump JuliaLowering: `macro_name`, `kw` changes

mlechu:
- `macro_name` is now a simple identifier, and `kw` is used in some kwarg cases.  Both changes should align with Expr.
- Remove `select_target_node` method on `SyntaxNode` (unused?) which is no longer the same as `SyntaxTree`, so would need a separate function if we wanted it.

I've pointed us to my own branch (https://github.com/JuliaLang/julia/pull/60327) so CI can run, but feel free to change that after it's done

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,8 +26,8 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 [sources]
 Glob = {rev = "avi/globstar", url = "https://github.com/aviatesk/Glob.jl"}
 JET = {rev = "f577f65", url = "https://github.com/aviatesk/JET.jl"}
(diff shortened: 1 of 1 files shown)


aviatesk (approved):
Thanks so much for working on this. The changes looks valid, great work.

aviatesk:
Confirmed CI runs successfully.
@mlechu You can just merge the Julia base PR yourself and delete the branch as usual. I will then take this PR and create a separate static branch in JuliaLang/julia.

mlechu:
Sounds good, I'll merge the base PR once CI is green there.  I think JuliaLowering and JETLS are the only users of the changed code anyway

PR #369, opened 2025-12-07
notebook: Implement Jupyter notebook document support

aviatesk:
https://github.com/user-attachments/assets/f7476257-7a53-44a1-8c8c-1ad57e136a63

Add support for Jupyter notebook documents in the language server, enabling features like completions, hover, definitions, rename, and diagnostics to work within notebook cells.

The implementation concatenates all code cells into a single virtual source document, allowing the existing analysis infrastructure to work unchanged. Each cell's content is joined with newlines, and `CellRange` records track the line offset where each cell begins in the concatenated source.

To bridge between cell-local positions (used by clients) and global positions (used internally), the following mapping functions are introduced:

- `adjust_position`: Converts cell-local position to global position by adding the cell's line offset
- `unadjust_position`/`unadjust_range`: Converts global position back to cell-local coordinates, also returning the cell URI

These mappings are applied at LSP request entry points (completions, hover, definition, rename, etc.) to adjust incoming positions, and when constructing responses to convert ranges back to cell coordinates.

For diagnostics, map_notebook_diagnostics! transforms diagnostics computed on the concatenated source into per-cell diagnostics by distributing them to the appropriate cell URIs based on line offsets.

Notebook document synchronization handles incremental text changes since `NotebookDocumentSyncOptions` does not provide an equivalent to `TextDocumentSyncKind.Full` for cell content.

jetls-client: Update vscode-languageclient to 10.0.0-next.18 to enable pull diagnostics support for notebook cells. This requires updating tsconfig.json to use node16 module resolution and es2022 target, and switching to `LogOutputChannel` as required by the new version.

LSP: Move `NotebookDocumentFilter` and `NotebookCellTextDocumentFilter` to basic-json-structures.jl to allow `DocumentSelector` to include notebook cell text document filters as per LSP 3.17.0 specification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Commit: [`HEAD`](https://github.com/aviatesk/JETLS.jl/commit/HEAD)
 - Diff: [`fd5f113...HEAD`](https://github.com/aviatesk/JETLS.jl/compare/fd5f113...HEAD)
 
+### Added
+
+- Jupyter notebook support: JETLS now provides language features for Julia code
+  cells in Jupyter notebooks, including completions, hover, go-to-definition,
+  rename, and diagnostics. All cells are analyzed together, so definitions from
+  earlier cells are available in later cells.
+
+  > JETLS × notebook LSP demo
+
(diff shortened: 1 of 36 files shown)


aviatesk:
https://github.com/user-attachments/assets/b5bb5201-d735-4a37-b430-932b519254ee

PR #378, opened 2025-12-09
Use Glob.jl v1.4

visr:
I got an error that the "avi/globstar" branch was deleted. Though I see as of 10 minutes ago we no longer need it. Nice work on https://github.com/vtjnash/Glob.jl/pull/50!


aviatesk:
Thanks!

Issue #389, opened 2025-12-13
Globals are suggested for auto-complete even for `foo.property|`


topolarity:
<img width="939" height="62" alt="Image" src="https://github.com/user-attachments/assets/4ffe8a30-f09e-4b6e-b1e7-b67b192f41b6" />

I don't think `globals` (or locals for that matter) are ever accessible via this syntax.

It'd be good not to pollute auto-complete with them.

aviatesk:
Yeah sounds reasonable.
We should probably disable global
completions either here when we don't know the exact value of `prefixtyp` https://github.com/aviatesk/JETLS.jl/blob/90119a0aeaff4075a7b523a56577a6ffc8ce0518/src/completions.jl#L224

PR #395, opened 2025-12-15
completions: disable completions when dot-prefix type is unknown

aviatesk:
Previously, dot-prefixed completions like `xarg.x` would show irrelevant
global completions when the prefix type could not be resolved.
For now, let's just disable completions entirely in such cases, to
provide a cleaner user experience by not showing unrelated suggestions.

Property completions for non-module values may be added in the future.

Fixes aviatesk/JETLS.jl#389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   arguments. For example, `func(a; kw) = a` now correctly reports `kw` as unused.
   Fixes aviatesk/JETLS.jl#390.
 
+### Changed
+
+- Completions now return no results when the prefix type is unknown.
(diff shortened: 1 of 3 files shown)


topolarity:
Thanks for the fix!

>  when the prefix type could not be resolved

Out of curiosity, when can the prefix type be resolved?

I haven't been able to find any case that JETLS was able to tell me the fields of a local variable or argument yet.

aviatesk:
> Out of curiosity, when can the prefix type be resolved?

The current implementation of `resolve_type` is quite limited and incorrect, as it only performs abstract interpretation for dot-prefixed expressions in global scope, so it can derive types only when the dot-prefixed object is a global constant.

However, by utilizing the `ASTTypeAnnotator` added in #227, it would become possible to perform type analysis including local variables.

PR #396, opened 2025-12-15
completions: add keyword completions

### LINKED ISSUE #386: `true` frequently auto-completes to `trues`
If you type `true<enter>` using JETLS, it frequently auto-corrects the line for you to `trues`:

Image

Compare this to an example from Typescript:

Image

which suggests the solution is to add the literals to the auto-complete suggestion list.

Since the pattern:
```julia
function foo(x)
    cond && return true
end
```
is pretty common, this is not hard to hit in practice.

aviatesk:
Add completion support for Julia keywords. Keywords are sourced from `REPL.REPLCompletions.sorted_keywords` and `sorted_keyvals`.

Keywords are given the lowest sort priority (`max_sort_text3`) so they should appear after local variables and globals. They are excluded from macro invocation and latex/emoji completion contexts.

Closes #386.


topolarity:
This feels much better, thanks!

PR #413, opened 2025-12-25
config: generate & check schema for configuration automatically

abap34:
## Summary

As the configuration has grown more complex, it has become increasingly difficult for users to write correct configuration files without tool support. This PR adds JSON Schema generation for the configuration to support user.

The schema is automatically generated at release time and published as a release asset, enabling:
- Autocompletion in editors that support schema-based TOML editing
- Type validation for configuration values

https://github.com/user-attachments/assets/4b5d297c-0f86-49e7-8ec5-8ee267966dcd

## Implementation details

For this feature, I developed [Struct2JSONSchema.jl](https://github.com/abap34/Struct2JSONSchema.jl), which generates JSON Schemas without requiring any modifications to existing struct definitions. This approach is well-suited for projects like this one, where schema generation is optional and should remain decoupled from the core codebase.

A new script `scripts/schema/generate.jl` uses Struct2JSONSchema.jl to generate a JSON Schema from the `JETLSConfig` struct definition. 

Users can reference the schema in their editor configuration:
```
https://github.com/aviatesk/JETLS.jl/releases/latest/download/jetls-config.schema.json
```

## Notes

- ~~Struct2JSONSchema.jl is not yet registered in the Julia package registry, so the release workflow currently references it directly from GitHub. Once it is registered, this will be switched to using the registry version. It would be also reasonable choice to wait for that registration before merging this PR. (This is why this PR is draft)~~ registered.
- After this feature has been merged and stabilized, I plan to submit the schema to SchemaStore. Once registered, most editors will be able to provide configuration support out of the box, without requiring users to manually configure a schema URL.

## Try it now

For testing purposes, a temporary schema is currently hosted at https://www.abap34.com/jetls-config.schema.json. You can try it by either:
- Configuring your editor's schema extension to use this URL, or
- Installing the taplo language server and adding the following line at the top of your `.JETLSConfig.toml`:
  ```toml
  #:schema https://www.abap34.com/jetls-config.schema.json
  ```

## Tasks

- [x] update CHANGELOG
- [x] add description
- [x] add default value

### DIFF
--- a/scripts/schema/Project.toml
+++ b/scripts/schema/Project.toml
@@ -8,4 +8,6 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 
 [sources]
 JETLS = {path = "../.."}
-Struct2JSONSchema = {rev = "fix-hash", url = "https://github.com/aviatesk/Struct2JSONSchema.jl"}
+
+[compat]
+Struct2JSONSchema = "0.5"


abap34:
Update:
Struct2JSONSchema.jl now supports adding descriptions to individual fields, which enables documentation to be shown on hover in configuration files.

image

Based on the discussion with @aviatesk, the ideal system we want to achieve is as follows:

1. Generate a schema from `JETLSConfig` (i.e. a plain Julia struct)
2. Distribute it as a JSON Schema
3. Also distribute it via `package.json`

Steps 1 and 2 have already been achieved.

To achieve step 3, there are still some remaining issues.

First, the `configuration` section of the `package.json` specification differs from standard JSON Schema.
For example, `$ref` and `definitions` cannot be used.
See:
https://code.visualstudio.com/api/references/contribution-points#c:~:text=Unsupported%20JSON%20Schema%20properties

However, this is easy problem, since it can be worked around by inlining references.

The biggest remaining task is support for default values.
Struct2JSONSchema.jl does not currently support default values.
More precisely, this can already be achieved via an override mechanism, but it is not exposed as a simple interface.

I plan to implement a straightforward API for registering default values, such as:
`register_default_value!(ctx, JETLSConfig, DEFAULT_CONFIG)`

In some cases, such as the `formatter` section, the hierarchy of the Julia structs differs from that of the resulting schema. Handling this requires additional user customization, but it should be feasible.

After that is done, I think it will be ready to merge once we finalize a small interface for how and where to define description and other data, and update the devdocs accordingly.

abap34:
I think it will take a few days to implement features related to default values, so please feel free to update the configuration structures as you like :)

Additional note: It would be nice if we could generate descriptions from docstrings, but unfortunately Configurations.jl does not seem to retain docstrings for individual fields. As we discussed before, this may require creating a separate package tailored for this purpose.

aviatesk:
Yeah, we need some macro hacks to preserve field docs information (just realized this when I were writing `LSP.@interface`), so it's not surprising if Configurations.jl doesn't implement it.

Anyway, the current `description.toml` approach should be fine. Once the default value work is done, I understand that when we add a new configuration, we'll need to edit these additional places:
- description.toml
- docs/src/configuration.md

And I understand that jetls-client/package.json (or part of it) will be  generated from `Struct2JSONSchema`.

abap34:
> I understand that when we add a new configuration, we'll need to edit these additional places...

Yes, that’s correct.
Would it be sufficient to document this workflow in DEVELOPMENT.md?

aviatesk:
Yes, this kind of information for developers should go to DEVELOPMENT.md.

abap34:
UPDATE:

Struct2JSONSchema.jl now supports features such as setting default values easily.
There are no remaining missing core features, and the merge should be soon I think.

Remaining tasks:
- Write developer documentation.
- Update the CI to also generate `package.json`.
- Update the implementation to align with the current configuration options.

aviatesk:
Great progress. Let me know if you need a new review round.

abap34:
Support for updating `jetls-client/package.json` is now working.

As shown below, `package.json` is updated automatically when `JETLSConfig` in `types.jl` is updated.
(The script execution part of the video is x10 sped up)

https://github.com/user-attachments/assets/8135678b-92ce-4bcd-8429-22fb2518e2d4


There is still one point that needs discussion.

* When and by whom should this process be run?
  * Currently, the schema file (`jetls-config.schema.json`) is generated and distributed by CI at release time and is not kept in the repository.
  * Updating `package.json` at release time in the same way seems natural.
  * However, if we want `master` to always remain in a runnable state, it may be necessary for CI to update it on each change.
    Since configuration changes are unlikely to be frequent, this is not expected to significantly clutter the Git history.

Once this point is decided, I will write the developer documentation, review description and some implementations, and then it should be ready to merge.

aviatesk:
Automatically updating package.json with bots or actions is one idea, but I'm concerned that it might make the operational policy a bit complicated.

Since config updates don't happen very frequently, it might be fine to force devs to update package.json whenever we make changes to configuration using the update script in ./scripts? And, on the jetls-client side, we would set up CI that verfies whether the currently checked-in package.json is consistent with the current JETLS configuration schema.
Regarding releases, I agree with you that preparing a script for release time distribution is a good idea.

abap34:
Taking your feedback into account, I changed the setup as follows.

* **On release**: generate the JSON schema and distribute it as an asset.
* **On push to `master`**: check whether there are any diffs.

I also rewrote `description.toml` more carefully and pushed the regenerated `package.json`.
While a large part of the diff comes from semantically equivalent changes such as field ordering, there were also semantic differences, including the following:

#### 1. Description updates

For example, the description of `debounce` was `Debounce time in seconds before triggering full analysis after a document change`, but I believe this should actually refer to saving. I tried to apply this kind of correction based on the docs as accurately as possible. However, there may still be mistakes in my understanding (and Claude’s, who helped me 😀), so I would appreciate it if you could review this part carefully and feel free to edit it.　

#### 2. Added fields

Fields such as `inlay_hint` and `latex_emoji` were missing in `package.json`, so they have been added.

.github/workflows/release.yml
```diff
@@ -154,6 +154,17 @@ jobs:
             echo 'EOF'
           } >> $GITHUB_OUTPUT
 
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: "1.12"
```
abap34: The version is fixed at 1.12. If you prefer a different specifier, please let me know.

schemas/description.toml
```diff
@@ -0,0 +1,53 @@
+[JETLSConfig]
+diagnostic = "Diagnostic configuration. See [Diagnostic configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/diagnostic)."
+full_analysis = "Configuration for full JET analysis. See [Full analysis configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/full_analysis)."
+testrunner = "TestRunner integration configuration. See [TestRunner integration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/testrunner)."
+formatter = "Formatter configuration. Can be a preset name (`Runic` or `JuliaFormatter`) or a custom formatter object. See [Formatter configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/formatter)."
+completion = "Completion configuration. See [Completion configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/completion)."
+code_lens = "Code lens configuration. See [Code lens configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/code_lens)."
```
abap34, line 7: The page　https://aviatesk.github.io/JETLS.jl/release/configuration/#config/code_lens does not exist yet.
The corresponding page currently exists at
https://aviatesk.github.io/JETLS.jl/dev/configuration/#config/code_lens .
I understand that this will be fixed once the release is made.

aviatesk: Yeah, I intentionally made this point to the not-yet-existing page. It will be generated once we make a new release.

.github/workflows/release.yml
aviatesk: 1.12 is fine for now.

aviatesk:
> Fields such as `inlay_hint` and `latex_emoji` were missing in `package.json`, so they have been added.

Thanks 👍

scripts/schema/update-pkg-json.jl
```diff
@@ -0,0 +1,58 @@
...
+
+if "--check" in ARGS
+    if settings["properties"] != expected_props
+        @warn "The properties in package.json do not match the expected schema. Please run this script without --check to update it."
+        exit(1)
+    end
+else
+    settings["properties"] = expected_props
+    open(package_json_path, "w") do io
+        write(io, JSON.json(package_json, 2))
```
aviatesk: Should we use `OrderedDict` or similar technique to ensure stable output? Or does JSON.jl already handle that functionality? I kinda want the output to be deterministic and ideally sorted alphabetically in terms of JSON's key order.

scripts/schema/generate.jl
```diff
@@ -0,0 +1,12 @@
...
+gen_ctx = SchemaContext()
+setup_ctx!(gen_ctx)
+
+schema = generate_schema(JETLS.JETLSConfig; ctx = gen_ctx)
+
+schemafile_path = joinpath(@__DIR__, "..", "..", "jetls-config.schema.json")
+open(schemafile_path, "w") do io
+    write(io, JSON.json(schema.doc, 2))
+end
+
```
aviatesk: I'm thinking of adding a `--check` option to this script as well.
The reason is, I want to add a feature like `jetls schema` to the main JETLS.jl app, so that JETLS can output the schema itself. For this, I'm thinking we could check in the JSON schema directly into the JETLS.jl git checkout, similar to `package.json` for jetls-client. The idea is that JETLS would then `include_dependency` the schema file and use it for the `jetls schema` app output.
Some editors can get JSON schemas directly from language server extensions and use them for autocompletion (https://github.com/zed-industries/zed/pull/48334). While editors might be able to pull schemas from a remote source, I think outputting it locally would be the best solution for consistency.

Moving back to the original point, for CI, similar to `package.json`, we could check if the currently checked-in schema is consistent with the JETLS state.

aviatesk:
Great job. Can we generate a similar schema for `InitOptions` too? We can work on that in a separate PR if you want though.

schemas/description.toml
```diff
@@ -0,0 +1,53 @@
...
+inlay_hint = "Inlay hint configuration. See [Inlay hint configuration](https://aviatesk.github.io/JETLS.jl/release/configuration/#config/inlay_hint)."
+initialization_options = "Static initialization options that are set once at server startup and require a server restart to take effect. See [Initialization options](https://aviatesk.github.io/JETLS.jl/release/launching/#init-options)."
```
aviatesk, line 9: I don't think the schema for `JETLSConfig` (, which basically corresponds to what handled by `workspace/configuration`) should not contain anything about `initialization_options`, since they are just different ways to configure the server.

scripts/schema/update-pkg-json.jl
abap34: As you pointed out, it seems necessary to make it deterministic.
Since computation time and implementation are not an issue, explicitly sorting the results before generating the output should be sufficient. I’ll add that.

schemas/description.toml
aviatesk: On second thought, maybe `description.toml` itself should be as is, as the single source for all descriptions of dynamic settings (`workspace/configuration`) and initialization options. Rather, JETLS should output three types of schemas, and CI should probably validate each of them:
- Schema for dynamic settings 
- Schema for initialization options
- Schema for `.JETLSConfig.toml` (schema of dynamic settings + initialization options)

scripts/schema/update-pkg-json.jl
aviatesk: That's good.

scripts/schema/setup-schema-context.jl
```diff
@@ -0,0 +1,105 @@
+using TOML
```
aviatesk, line 1: We might want to make this script's name a bit clearer. Maybe something like `setup-context.jl`, or even `setup-schema-context.jl` – even though there is a duplication with the folder name. No strong preference, so you decide.

aviatesk:
Working beautifully. Thanks so much for your work on this, @abap34 !
Screenshot 2026-02-25 at 20 10 53

PR #437, opened 2026-01-04
completion: Use call-site types for return type inference

aviatesk:
Store `Core.MethodMatch` objects directly in the resolver info instead of navigating through modules to look up methods during resolution. The completion data now contains just a resolver ID and match index, which is used to retrieve the original match when resolving.

This simplifies the code significantly by removing `prepare_method_signature_data`, `lookup_method_from_data`, and `method_sparams`. The new `infer_match!` function uses `CC.specialize_method(match)` directly, which provides more precise return type inference since `MethodMatch` contains the actual argument types from the call site rather than the generic method signature types.

Also changes the return type display from `::Type` to ` -> Type` for better readability.

> Before
Screenshot 2026-01-04 at 18 48 22

> After
Screenshot 2026-01-04 at 18 47 17

### DIFF
--- a/LSP/src/language-features/completions.jl
+++ b/LSP/src/language-features/completions.jl
@@ -229,13 +229,12 @@ end
 
 # JETLS specific data structures for `data` field of `CompletionItem`
 struct GlobalCompletionData
+    resolver_id::String
     name::String
 end
 struct MethodSignatureCompletionData
-    moduleidx::Int
-    names::Vector{String}
-    methodname::String
-    methodidx::Int
+    resolver_id::String
+    match_idx::Int
 end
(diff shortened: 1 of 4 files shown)


topolarity:
> Also changes the return type display from ::Type to -> Type for better readability.

FWIW the guidelines for doc-comments in Base distinguish between these two. IIRC `->` is used to give a name to something or destructure it

For example `write(io::IO, string::String) -> bytes_written` or `split_at(...) -> (head::String, tail::Vector{String})`

aviatesk:
Ok, let's revert it to ` ::Ty` then.

aviatesk:
On VSCode, it would appear as following:
Screenshot 2026-01-04 at 23 27 18

PR #440, opened 2026-01-04
completions: Change type annotation syntax from `->` to `::`

aviatesk:
Updates the completion detail format to use Julia's type annotation syntax `::` instead of function return type syntax `->` for better consistency with Julia conventions.

xref:
https://github.com/aviatesk/JETLS.jl/pull/437#issuecomment-3708079156

> Before (VSCode)
Screenshot 2026-01-04 at 23 30 20

> After (VSCode)
image

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -784,7 +784,7 @@ function resolve_completion_item(state::ServerState, item::CompletionItem)
         _, result = infer_match!(CC.NativeInterpreter(Base.get_world_counter()), match)
         rettyp = CC.widenconst(result.result)
         # TODO Show effects and exception type?
-        detail = " -> " * completion_resolver_info.postprocessor(string(rettyp))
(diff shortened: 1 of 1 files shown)


topolarity:
Uh oh - That VS Code and other editors don't agree w.r.t. how to display this is a bit unfortunate

If we were dealing with only VS Code, I might prefer `returns Union{Float64, Int64}` or just `Union{Float64,Int64}` on its own. But given that this is concatenated in other editors, I guess we have to make a compromise

aviatesk:
Yea, LSP doesn't precisely define how this information should be displayed. For my personal taste, I like `-> Union{Float32, Float64}` the most, but a conflict with the Julia documentation policy would be an issue. On the other hand, if we just show return type name, it could become very unclear in other editors. I think `::` is the best compromise.

topolarity:
That's fair - Given the extra details, I'd also be fine with either `->` or `::`

PR #470, opened 2026-01-10
Use video elements on notebook docs

visr:
When going to this pages it starts to download the videos to disk instead of embedding it: https://aviatesk.github.io/JETLS.jl/dev/notebook/

image

This PR fixes that:

image


aviatesk:
Thanks!

PR #472, opened 2026-01-14
Correct configuration schema

nandoconde:
Correct schema for config file, formatter name is supposed to be outside any section.

image


aviatesk (approved):
Thanks!

PR #483, opened 2026-01-16
Use `"jetls"` as namespace for `workspace/configuration`

### LINKED ISSUE #482: Configuration schema incompatible with neovim
JETLS's `workspace/configuration` request sets both `scopeUri = null` and `section = null`, and expects to receive the full config in response. However, different clients interpret `section = null` differently, and this approach is incompatible with some of them. In particular, neovim will return an empty response, so there's no way to configure JETLS from neovim.[^1]

A fairly common pattern to ensure compatibility is to namespace the entire configuration under a single server-specific section. This is used by rust-analyzer, LuaLS, pyright/basedpyright, texlab, and markdown_oxide, just to name a few.

So instead of `section = null`, JETLS would use, say, `section = "jetls"`. The neovim client would then be configured as follows (here I'm combining the docs' example configuration with the basic server setup from the index page):

```lua
vim.lsp.config("jetls", {
  cmd = {
    "jetls",
    "--threads=auto",
    "--",
  },
  filetypes = { "julia" },
  root_markers = { "Project.toml" },
  settings = {
    jetls = {
      full_analysis = {
        debounce = 2.0,
      },
      -- Use JuliaFormatter instead of Runic
      formatter = "JuliaFormatter",
      diagnostic = {
        patterns = [
          -- Suppress toplevel/inference warnings in test folder
          {
            pattern = "(toplevel|inference)/.*",
            match_by = "code",
            match_type = "regex",
            severity = "off",
            path = "test/**/*.jl",
          },
        ],
      },
      testrunner = {
        executable = "/path/to/custom/testrunner",
      }
    },
  },
})
```

This is a trivial change on the server side, just changing `nothing` -> `"jetls"` in a single location, and I've confirmed locally that it works.

Another alternative, which would preserve the current config schema, is to instantiate a separate `ConfigurationItem` for each available config key; that is, one item with `section = "full_analysis"`, one with `section = "formatter"`, and so on.

Either way, using non-null, non-empty values for `section` would make JETLS work better with neovim, and improve compatibility across clients in general. If you approve of one or the other solution I'd be happy to make the PR.

[^1]: Just looking at the code, it looks as if neovim would return the whole config if you set `section = ""` (empty string). See https://github.com/neovim/neovim/blob/v0.11.5/runtime/lua/vim/lsp/handlers.lua#L187-L198. However, I couldn't get this to work. Non-empty keys appear to be the only safe bet.

danielwe:
Fixes #482.

Also makes it so that empty strings are no longer omitted, i.e., equivalent to no value, when encoding JSON messages. Omitting them is why neovim wasn't returning any configuration even when I tried to use `section = ""`---neovim was still just seeing `nil` on it's end. (UPDATE: The PR now uses `"jetls"` instead of `""` as the top-level config section, see discussion below, but I'm keeping the change to empty string handling as it seems like the right thing for correctness.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 `textDocument/rename`, and `textDocument/definition` by avoiding re-parsing of
   already analyzed files not opened in the editor.
 
+- `workspace/configuration` requests now expect settings to be found under the
+  top-level `"jetls"` key, such that a request with `section = "jetls"` produces
+  the full configuration. This is to ensure compatibility with generic clients,
+  e.g., the neovim client, which may not conform to JETLS's previous
+  expectations about how requests with no `section` are handled.
+
 ### Fixed
 
 - Fixed LSP features not working inside `@main` functions.
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -552,6 +552,42 @@ section:
 }
 ```
 
+#### Server-agnostic clients (e.g., neovim, emacs lsp-mode, helix)(@id config/lsp-config/server-agnostic)
+
+Settings should be placed under the `"jetls"` key, such that a request for the
+`"jetls"` section produces an instance of the JETLS configuration
+[schema](@ref config/schema). For example, neovim's built-in LSP client may be
+configured as follows:
+
+```lua
+vim.lsp.config("jetls", {
+  settings = {
+    jetls = {
+      full_analysis = {
+        debounce = 2.0,
+      },
+      -- Use JuliaFormatter instead of Runic
+      formatter = "JuliaFormatter",
+      diagnostic = {
+        patterns = [
+          -- Suppress toplevel/inference warnings in test folder
+          {
+            pattern = "(toplevel|inference)/.*",
+            match_by = "code",
+            match_type = "regex",
+            severity = "off",
+            path = "test/**/*.jl",
+          },
+        ],
+      },
+      testrunner = {
+        executable = "/path/to/custom/testrunner"
+      },
+    },
+  },
+})
+```
+
 ## [Configuration priority](@id config/priority)
 
 When multiple configuration sources are present, they are merged in priority
--- a/src/workspace-configuration.jl
+++ b/src/workspace-configuration.jl
@@ -1,3 +1,5 @@
+const WORKSPACE_CONFIGURATION_NAMESPACE = "jetls"
+
 struct LoadLSPConfigHandler
     server::Server
     source::String
@@ -50,7 +52,7 @@ end
 
 function load_lsp_config!(server::Server, source::AbstractString; on_init::Bool=false)
     handler = LoadLSPConfigHandler(server, source, on_init)
-    request_workspace_configuration(handler, server, nothing)
+    request_workspace_configuration(handler, server, WORKSPACE_CONFIGURATION_NAMESPACE)
     nothing
 end
 


jwortmann:
Hello,

note that while an empty string is technically allowed, I doubt that this is intended or would be a useful as an actual value for `section`. VSCode and also Sublime Text use settings files in JSON format and allow to define the settings as namespaced (dotted) strings; for example
```json
{
    "full_analysis.debounce": 2.0
}
```

For the response this gets automatically converted to a nested JSON object `"full_analysis": {"debounce": 2.0}` by the clients.

If you now use an empty string for `section`, users would be required to define the settings like
```json
{
    ".full_analysis.debounce": 2.0
}
```
which is kinda awkward and might even result in buggy behavior in case the client doesn't expect such empty namespaces (`section`).

Note that contrary to what was claimed in https://github.com/aviatesk/JETLS.jl/issues/482#issue-3819041082, JETLS *does not* send `section = null` (that would in fact be a spec violation, because `section` must be a string). Instead, it simply omits the optional `section` key, which is valid per LSP specs. It is up to the client then what to query from the settings file and to send back in the response. 

---

I think a feasible solution would be to use `"jetls"` as the value for `section` as you suggested in #482. Ideally such a change should be mentioned in the changelog / release notes, because it would require users to adjust their setting keys to e.g. `"jetls.full_analysis.debounce"`.

danielwe:
> I doubt that this is intended

I don't know about the protocol, but in the neovim client, the treatment of `""` is explicit and intentional, as you can see here: https://github.com/neovim/neovim/blob/v0.11.5/runtime/lua/vim/lsp/handlers.lua#L187-L198. Because of the special casing, it does not require the leading period you suggest. Given
```lua
{ full_analysis = { debounce = 2.0 } }
```
the responses will be
```
""                       -> {"full_analysis": {"debounce": 2.0}}
"full_analysis"          -> {"debounce": 2.0}
"full_analysis.debounce" -> 2.0
```
But from what you're saying it sounds like VSCode and Sublime behave differently and wouldn't handle this as gracefully. If that's the case, this is obviously a non-starter.

Using `""` seemed like the least intrusive fix here, since it wouldn't require adding a nesting level in the configuration schema. But I suppose that's not really an important concern for software that's still on version 0.0.0. If "jetls" is the preferred solution, I'd be happy to redo the PR accordingly.

> JETLS _does not_ send `section = null`

You're right of course. As you can see further down the thread in https://github.com/aviatesk/JETLS.jl/issues/482#issuecomment-3757739421, my understanding gradually improved as I peeled back more layers. (I guess the distinction is important for a typescript endpoint where `null` and `undefined` are distinct concepts. In the case I've been debugging it's Julia talking to Lua, so you just end up with `nothing`/`nil` either way.)

Irrespective of how the JETLS config is best structured, I'm curious: What is the correct way to encode an empty string per the spec? Omit as before, or keep as in this PR?

jwortmann:
I took a second look at the client implementation in Sublime and saw that actually the empty string would work there without requiring another nested level in the configuration as well: https://github.com/sublimelsp/LSP/blob/ea0963d8ba474f47ba41fbc5999d0c84d933e56e/plugin/core/sessions.py#L2166
Here the `get('section') or None` would treat an empty string exactly as if `section` would not be present in the [ConfigurationItem](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#configurationItem) because `bool("")` evaluates to `False` in Python. Nevertheless, I have not tested whether it works in VSCode and I would still not recommend to use an empty string for `section`, since such a behavior is not really declared in the LSP specs and other clients might handle it differently.

---

> What is the correct way to encode an empty string per the spec?

I don't have a concrete example where an empty string would be useful (typically properties which are not strictly necessary are declared as optional keys in the JSON/TypeScript definitions from the LSP specs), but also there shouldn't be any reason to disallow empty strings. So I would say that part of this PR is fine and they should probably not be omitted. But also I don't know if anywhere else in the JETLS code base an empty string is used intentionally to omit the corresponding property in the JSON payload, so this should probably be checked by the maintainer to ensure it doesn't break anything.

danielwe:
I pushed an update to use `"jetls"` as the toplevel key. Also updated the changelog, and added a section in the docs for configuring JETLS via neovim/generic clients.

I didn't change the docs for VSCode and Zed for now, as I reckon there are several ways to resolve this change on their end. They could just update their internal mapping of requests such that users don't have to modify their editor settings, or they too could insert the `"jetls"` namespace in their config schemas for more of a one-to-one mapping and perhaps less confusion for users exploring multiple clients. I don't have an opinion, as I don't use either.

aviatesk:
Thanks for working on this.

I have been confused that `workspace/configuration` is pretty underspecified (as discussed at microsoft/language-server-protocol#972), and different clients have ended up with different conventions for what “root” means.

Still I’d prefer that we don’t require users to introduce an extra top-level `"jetls"` namespace in their client settings.. E.g. in your docuemntation example, settings is already scoped per-server by the client configuration, so forcing `settings = { jetls = { ... } }` would effectively double-namespace and feels unnatural to me.

Instead, I think the simplest server-side approach without `clientInfo` branching would be to make the `workspace/configuration` request tolerant by including two `ConfigurationItems` in a single request?:
- one with `section` omitted / null
- one with `section = ""`

and then select the first non-null result. That keeps the interaction generic, avoids special-casing specific clients, and avoids introducing a mandatory `"jetls"` prefix just to satisfy one client’s convention. (and it's probably better to state this in our documentation)

Regarding the serialization change: I think we should scope it narrowly. For this PR, please apply the “do not omit empty strings” behavior only for `ConfigurationItem` so we don’t accidentally change semantics for other interfaces. If we want to generalize this across all `@interfaces`, we can do that in a separate PR.

danielwe:
> forcing `settings = { jetls = { ... } }` would effectively double-namespace and feels unnatural to me.

I agree that it feels redundant, but as I mentioned in https://github.com/aviatesk/JETLS.jl/issues/482#issue-3819041082, it seems to have emerged as a widespread convention to avoid the underspecified edge cases discussed in https://github.com/microsoft/language-server-protocol/issues/972. (I was also reading that thread when debugging this.)

Here's a small and somewhat editorialized snapshot of my neovim config:
```lua
{
  "neovim/nvim-lspconfig",
  opts = {
    servers = {
      basedpyright = {
        settings = {
          basedpyright = {
            analysis = {
              typeCheckingMode = "standard",
            },
          },
        },
      },
      julials = {
        settings = {
          julia = {
            completionmode = "qualify",
          },
        },
      },
      texlab = {
        settings = {
          texlab = {
            inlayHints = {
              maxLength = 13,
            },
          },
        },
      },
      lua_ls = {
        settings = {
          Lua = {
            codeLens = {
              enable = true,
            },
          },
        },
      },
    },
  },
}
```

And here's from my brief experiment with helix a little while ago:
```toml
# ~/.config/helix/languages.toml
[language-server.texlab.config.texlab]
build = { onSave = true, forwardSearchAfter = true }
forwardSearch.executable = "/Applications/Skim.app/Contents/SharedSupport/displayline"
forwardSearch.args = ["-r", "-g", "%l", "%p", "%f"]
``` 

I don't use Rust, but as far as I understand, `rust-analyzer` expects the same pattern.

I don't _like_ the deep nesting and redundant namespacing, but it looks like a lot of LSP authors have converged on this solution to ensure compatibility and robustness across as many different editors/clients as possible.

> avoids special-casing specific clients, and avoids introducing a mandatory `"jetls"` prefix just to satisfy one client’s convention.

I see this as the opposite: steering clear of the underspecified edge cases is how you avoid as much as possible the need for special-casing. This isn't a neovim-specific convention, the point is to ensure that JETLS can communicate with _any_ client without issue, and without having to figure out editor-specific tricks and tweaks to the client-side config.[^1]

---

All that said, I'm just a drive-by, first-time contributor, so it's time for me to step down from my soapbox. I'll implement whatever version you prefer, or feel free to edit this PR yourself, or close it in favor of another solution. I'm happy either way!

But if we go for the solution you suggest, it would be good to confirm with as many editors as possible that at least one of the two items in the request produces the response we want. I can check helix in addition to neovim.

jwortmann:
I assume this convention with the non-empty namespaces like "jetls" originates from VSCode using a single `settings.json` configuration file with all settings included at the top level JSON object. For example, your neovim config would correspond to
```json
{
    "basedpyright.analysis.typeCheckingMode": "standard",
    "julia.completionmode": "qualify",
    "texlab.inlayHints.maxLength": 13,
    "Lua.codeLens.enable": true
}
```

These namespaces typically are the same as the extension name or language name, since the language servers are often bundled with the language extension in VSCode (like the Julia extension with LanguageServer.jl). Here such a namespace prefix is necessary to prevent clashes of the same setting names from different extensions.

In Sublime Text this is not really necessary because that editor uses individual setting files for each extension (alternatively the server config can also be declared in a shared file similar to the neovim config, in case there exists no standalone extension package for the language server). So there is no actual need of the non-empty namespace prefix for that client, but due to using the dotted strings in JSON format the prefix also doesn't really add any complexity.

But note that the LSP specs also has an explicit example for the `section` value in the configuration request: `cpp.formatterOptions`
(see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration)
So if you want to follow the common convention, I'd say "jetls" would be more appropriate than an empty string or even duplicated ConfigurationItems in the request params.

danielwe:
There are a couple of different things to keep in mind here:

1. What keys may the server request?
2. What structure does it expect in response to each key?
3. How are settings structured and represented on the client side?
4. How is a key translated to a selection of data from the client-side settings store, and from there to an appropriately structured response to the server?

As far as I can tell, for VSCode (and maybe Zed? Sublime?), the norm is to supply a tailor-made client plugin for each server. This provides a lot of flexibility wrt. point 4. The client-side settings may require the namespacing you describe, but the server can still send empty requests if that's what the devs decide. The client plugin will just have to translate back and forth.

In neovim and helix, client configuration is mostly declarative and rarely relies plugins (helix doesn't even have a plugin system), so there's no way to customize point 4.[^1] You simply provide a table of settings knowing that the client will interpret `section` as a literal key into this table, with support for dotted strings to retrieve nested subtables. Send an empty request, get an empty response. Send `section="foo"`, receive `<servername>.settings["foo"]`. As we've discussed, the only special case in neovim is that `section=""` will return `<servername>.settings` (unless `<servername>.settings[""]` exists). I haven't checked if Helix implements the same special case.

As a result, there are three options for structuring and requesting your configuration:

1. Don't use a namespace; directly request the subtables you're interested in by sending separate `ConfigurationItem`s for `section="formatter"`, `section="full_analysis"`, et cetera.
2. Don't use a namespace; rely on the `section=""` special case to retrieve the full configuration in one item. Cross your fingers that this will keep working and become an established norm among the more declaratively configured clients.
3. Use a namespace; use this key, e.g., `section="jetls"`, to unambiguously request the full configuration in one item.

Most of the servers I come across seem to prefer 3, though some seem to use 1 (e.g., tinymist, unless I'm mistaken). I'm sure there are servers out there who do 2, otherwise neovim probably wouldn't have the special case, but I don't think I've come across one.

What I'd like to emphasize is that if you choose an approach that works reliably under these constraints (that is, 1 or 3), it's easy to make it work in any context, including in the more plugin-heavy editors. The converse does not hold. Hence my claim that these approaches do not entail special-casing for one client, but rather the opposite.

---

Once again, I'm aware that I'm coming in hot and full of opinions in my first time interacting with this project, so now I'll slow down, stop soapboxing for real, and await the conclusion. Thanks for a great and promising project, I'll be a happy user regardless of the outcome here!

[^1]: Unless you write your own LSP client from scratch and sidestep the built-in one.

danielwe:
Per the source, it looks like helix also supports the empty string for `section`, even if they're somewhat bewildered by the necessity. https://github.com/helix-editor/helix/blob/99ebccbfc5f6b414e442e72ca6ce4fd75d2d8f0b/helix-term/src/application.rs#L1035-L1040

Looks like their implementation also supports omitting `section`.

aviatesk:
Apologize for the delay in feedback and thank you again for investigating this issue.
I have reviewed the discussion results again, and I conclude that, as you suggested, always requesting with the `"jetls"` prefix would be the most practical and best approach. While this might require redundant settings for some clients, it stems from the ambiguity in the LSP specification, and given the current specification, there are unavoidable aspects.

As I believe you are already aware, language servers are primarily used in two ways:
- where a dedicated language client extension is prepared on the client side: e.g. Zed, VSCode
- where a server-agnostic language client is provided with the editor: Neovim, Helix, Emacs, etc.

In the former case, we would need our own language client implementation anyway. So regardless of how JETLS specifies `section` in that client, the client extension can freely handle it appropriately and return settings in the desired format for that client, meaning it is not affected by this PR.

The latter case is where the absence of the `"jetls"` prefix becomes an issue. More precisely, it's problematic in server-agnostic language client implementations where the client's configuration does not have a pre-configured dedicated setting section for a specific server, and instead relies on the LSP's `section` to return server-specific settings.

For example, consider an agnostic client implementation that uses the following configuration style:
```json
{
    "servers": {
        "JETLS": {
            "binary": "jetls",
            "settings": {
                ...
            }
        }
    }
}
```

One reasonable way to implement `workspace/configuration` in such a client is to automatically return `servers.JETLS.settings` if the `workspace/configuration` request came from JETLS (let's call this client "A"). In such cases, specifying `section` would likely not be necessary.

However, in LSP, `workspace/configuration` is written such that a server can request settings that are not necessarily specific to that server, such as `editor.indent` (this ambiguity is precisely the problem though). Therefore, I believe many clients do not adopt "A", but instead settle on an implementation (let's call this "B") where they return global settings even when JETLS sends a `workspace/configuration` request.

In such clients with the "B" style, the `"jetls"` prefix is necessary to distinguish from other servers:
```json
{
    "servers": {
        "JETLS": {
            "binary": "jetls",
            "settings": {
                "jetls": {
                    "diagnostic": {
                        "all_filse": false
                    }
                }
            }
        }
    }
}
```

I felt that the `jetls` part in this `"servers.JETLS.settings.jetls"` path was redundant, and I didn't want to require it even for "A" clients. However, for the "B" clients, `"section = jetls"` is essential, so it seems unavoidable. This is not because "B" clients are 'bad,' but rather because they are trying to be as faithful as possible to the ambiguous LSP specification.

Based on your research, I think Neovim and Helix are trying to support both of these approaches IIUC? They seem to implement it by basically taking settings from the global settings space when a `section` is specified, while special-casing `""` to return server-specific settings only in that case. (they behave like B when `section` is specified or `null`, and like A when `section == ""`.)

Well, in any case, I think adding `"jetls"` would be the best way to go. I can tweak the existing dedicated language client extensions for VSCode and Zed later if necessary.

> What keys may the server request?

Regarding this point, JETLS will continue to use the method of requesting "all JETLS-related sections" in the future. That is, it will not request `section` with `"jetls.full_analysis"` or `"full_analysis"`. Our configuration handler implementation always expects to receive a full JETLS configuration section object. This approach is simpler and does not pose performance issues. (I've thought about requesting items like `editor.indent` for various LSP features. But considering how `workspace/configuration` is currently designed, it would probably be impossible to implement this in a client-agnostic way. That's why I'm frustrated that LSP implies you can request `editor.indent` through `workspace/configuration`...)

LSP/src/DSL/interface.jl
```diff
@@ -172,6 +172,8 @@ function process_interface_def!(toplevelblk::Expr, structbody::Expr,
             # (except in the tests)
             omitempties = ()
         end
+        # Empty strings are valid data and should not be omitted for being empty containers
+        push!(toplevelblk.args, :(StructTypes.isempty(::Type{$Name}, ::AbstractString) = false))
```
aviatesk: Let's move this change into a separate PR.

CHANGELOG.md
```diff
@@ -56,6 +56,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   the container name for arguments or fields respectively, making it clearer
   which function or struct they belong to during workspace symbol search.
 
+- `workspace/configuration` requests now expect all settings to be found under
+  the top-level `"jetls"` key, such that a request with `section = "jetls"`
+  produces the full configuration. This ensure compatibility with generic
+  clients such as neovim's, which may not conform to JETLS's previous
+  expectations about how requests with no `section` are handled.
```
aviatesk: 👍 Good to have a CHANGELOG entry for this.

danielwe:
Sounds good! I've rebased, removed the empty string parts (never mind the branch name...), and improved the wording in the changelog and docs.

> Based on your research, I think Neovim and Helix are trying to support both of these approaches IIUC? They seem to implement it by basically taking settings from the global settings space when a `section` is specified, while special-casing `""` to return server-specific settings only in that case. (they behave like B when `section` is specified or `null`, and like A when `section == ""`.)

Sort of... They're not 100 % consistent. When `section` is omitted (null), neovim returns an empty response, while helix returns the equivalent of `servers.JETLS.settings`, i.e., behavior A in your taxonomy. Neovim's empty response is why it does not work with JETLS as-is. Helix, on the other hand, should work fine (just based on reading code; I haven't tested).

Both editors special-case `section = ""` to mean behavior A. The reason for this seems to be to appease certain servers, such as ESLint, that expect this. Based on @jwortmann's research it seems like Sublime does the same.

What every client seems to agree on is that `section = "jetls"` should return the equivalent of `servers.JETLS.settings.jetls`. Just like you, I don't _like_ the redundancy of this, but it does appear to be the only choice that avoids ambiguous edge cases of the spec and thus ensures compatibility with every kind of client. Hence why so many other servers have also settled for this approach.

Yesterday, I noticed that neovim has an open PR that would make it work like helix when `section` is omitted: neovim/neovim#27510. I think this would be a good move, and if it had already been merged, this whole discussion/PR would might never have happened. But the fact that it hasn't is evidence that servers ought to avoid these underspecified edge cases.

PR #493, opened 2026-01-17
Fix instructions for eglot configuration

giordano:
You can't just add to a non-existing list in the top-level, you only get an error.  `add-to-list` must be wrapped in a macro which defers its evaluation, like `with-eval-after-load`.  That's also the recommended syntax in [Eglot documentation](https://elpa.gnu.org/devel/doc/eglot.html#Setting-Up-LSP-Servers-1) (and really the standard approach when configuring Elisp packages)


aviatesk:
Thanks!

PR #314, opened 2025-11-22
release: Implement environment isolation via dependency vendoring

aviatesk:
This commit introduces a release process that isolates JETLS's dependencies from packages being analyzed by vendoring them with rewritten UUIDs.

The vendor-deps.jl script automates the entire release process:
1. Fetches clean Project.toml from the development branch
2. Updates and vendors all non-stdlib, non-JLL dependencies
3. Rewrites UUIDs deterministically using uuid5
4. Updates Project.toml with vendored dependencies and `[sources]` entries

This approach ensures JETLS's dependencies never conflict with packages users are analyzing, resolving issues where version conflicts between JETLS and analyzed packages would prevent analysis.

---

The following describes the current release process using the above script (at least until JETLS has an official release process).

Fow now, we employ the following branch strategy:
- Development branch (`master`): Regular development with normal UUIDs
- Release branch (`release`): Vendored dependencies with rewritten UUIDs

Then the `release` branch is managed as follows:
1. Check out the `release` branch
2. Merge (not rebase) changes from the development branch: `git merge -X theirs master`[^merge-strategy]
3. Vendor dependency packages: `julia vendor-deps.jl --source-branch=master`
4. Commit changes
5. Push to remote `release`

[^merge-strategy]: The `-X theirs` option automatically resolves merge conflicts by preferring the `master` branch version. This is necessary because `Project.toml` files are always modified on the `release` branch (with vendored UUIDs), causing conflicts when merging. Since `vendor-deps.jl` fetches and overwrites `Project.toml` from `master` anyway in step 3, these conflicts can be safely resolved by taking the `master` version. Using `git merge` (rather than `git rebase`) keeps the commit history linear for users who update via `git pull`.

This means users would manually run `git pull origin release` followed by `Pkg.update()` to install and update JETLS (this is almost the same as the  current process, except that we're pulling from `master`). However, I may change this distribution method to make it easier using Pkg apps.

For detailed consideration of this vendor approach and release process, refer to:
- https://publish.obsidian.md/jetls/work/JETLS/Release+process+for+JETLS+alpha
- https://publish.obsidian.md/jetls/work/JETLS/JETLS+environment+isolation

PR #499, opened 2026-01-18
fix: unclosed test runner

MilesCranmer:
Small change to wrap some blocks with `try {...}; finally; close(testrunnerproc); end`. Otherwise it might keep running in the background.

### DIFF
--- a/src/testrunner/testrunner.jl
+++ b/src/testrunner/testrunner.jl
@@ -528,7 +528,6 @@ function _testrunner_run_testset(
     finally
         close(testrunnerproc)
     end
-    result isa AbstractString && return result
 
     ret = summary_testrunner_result(result)
 
@@ -657,7 +656,6 @@ function _testrunner_run_testcase(
     finally
         close(testrunnerproc)
     end
-    result isa AbstractString && return result
 
     # Show the results of this `@test` case temporarily as diagnostics:
     # The `Server` (or `FileInfo`) doesn't track the state of each `@test`,


src/testrunner/testrunner.jl
```diff
@@ -501,74 +501,78 @@ function _testrunner_run_testset(
     cmd = testrunner_cmd(executable, filepath, tsn, tsl, test_env_path)
     testrunnerproc = open(cmd; read=true)
 
-    # Wait for the process with cancellation support
-    while true
-        process_running(testrunnerproc) || break
+    try
```
aviatesk: Should we limit the scope of this `try` block to the lifetime of `testrunnerproc`, i.e.
```julia
    result = try
        LSP.JSON3.read(testrunnerproc, TestRunnerResult)
    catch err
        @error "Error from testrunner process" err
        show_error_message(server, """
            An unexpected error occurred while executing TestRunner.jl:
            See the server log for details.
            """)
        return "Test execution failed"
    end
```
?

aviatesk (commented):
Thanks, I think this is a good change. 
Have you run into a case where `testrunnerproc` wasn't killed? If so, there might also be an issue on the TestRunner side. But we haven't implemented [environment isolation](https://github.com/aviatesk/JETLS.jl/pull/314) for TestRunner yet, so issues stemming from that are a known problem. So this `try/finally` block is good to have anyway.

MilesCranmer:
I have seen some instances where there's a Julia process that runs in the background and burns a lot of CPU% long after I've stopped working on something in VSCode. However I do not know if that was from this specific branch. But looking through, I noticed it, so thought we might as well patch it just in case.

aviatesk:
Ok. It's fine to look into deeper whenever we re-encounter the issue then. I'd like to merge this after minimizing the scope of `try`/`finally`.

src/testrunner/testrunner.jl
```diff
@@ -501,29 +501,35 @@ function _testrunner_run_testset(
...
             An unexpected error occurred while executing TestRunner.jl:
             See the server log for details.
             """)
-        return "Test execution failed"
+            return "Test execution failed"
+        end
+    finally
+        close(testrunnerproc)
     end
+    result isa AbstractString && return result
```
aviatesk: Why do we need this check?

PR #533, opened 2026-01-31
docs (mostly): place subcommand `serve` after `--` separator

danielwe:
With the introduction of subcommands, the docs on editor setup generally suggest the following launch command: `jetls serve --threads=auto --`. But this command errors out, as `serve` is interpreted by the Julia process as the name of a file to execute. App args like `serve` must come after the `--` separator, while Julia args like `--threads` must come before, so the correct invocation is `jetls --threads=auto -- serve`.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -96,9 +96,9 @@ Minimal [Emacs](https://www.gnu.org/software/emacs/)
                '(((julia-mode :language-id "julia")
                   (julia-ts-mode :language-id "julia"))
                  "jetls"
-                 "serve"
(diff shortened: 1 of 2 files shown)


aviatesk:
You're right, thanks for fixing this up.
Since we're setting `julia_flags`, strictly speaking these runtime flags might not be necessary, but well, let's check that later and cleanup if needed.
https://github.com/aviatesk/JETLS.jl/blob/76f986ed2699219f639784f26aa49a6610f74ef8/Project.toml#L51

Issue #481, opened 2026-01-15
Incorrect unused keyword argument in functions


avik-pal:
```julia
function compute_rotary_embedding_params(
    head_dim::Integer,
    max_sequence_length::Integer;
    base::Number,
    dtype::Type{T}=Float32,
    low_memory_variant::Bool=true,
) where {T}
    θ = inv.(T.(base .^ (range(0, head_dim - 1; step=2)[1:(head_dim ÷ 2)] ./ head_dim)))
    seq_idx = collect(T, 0:(max_sequence_length - 1))
    angles = reshape(θ, :, 1) .* reshape(seq_idx, 1, :)
    low_memory_variant || (angles = vcat(angles, angles))
    return (; cos_cache=cos.(angles), sin_cache=sin.(angles))
end
```

`dtype` shows up as unused (while `T` is used)

PR #545, opened 2026-02-14
diagnostic: Fix false positive unused kwarg with where clause

aviatesk:
Fix incorrect "unused argument" diagnostic for keyword arguments whose type annotation constrains a `where`-clause static parameter that is used in the function body. For example, in:

  `f(; dtype::Type{T}=Float32) where {T} = T.(xs)`

`dtype` was incorrectly reported as unused even though it provides `T` which is used. This exception only applies to keyword arguments since positional arguments can use `_::Type{T}` or `::Type{T}`.

The fix pre-computes a mapping from source locations to identifier names in keyword argument type annotations by traversing `st0` once, then checks if any of those names match a `:static_parameter` binding with `is_read == true`.

Closes aviatesk/JETLS.jl#481

### DIFF
--- a/src/diagnostic.jl
+++ b/src/diagnostic.jl
@@ -655,7 +655,7 @@ function analyze_unused_bindings!(
         provs = JS.flattened_provenance(JL.binding_ex(ctx3, binfo.id))
         is_from_user_ast(provs) || continue
         prov = last(provs)
-        if bk === :argument && kwarg_type_names !== nothing
+        if bk === :argument
             prov_loc = JS.source_location(prov)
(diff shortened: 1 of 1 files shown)


src/diagnostic.jl
```diff
@@ -602,6 +655,12 @@ function analyze_unused_bindings!(
         provs = JS.flattened_provenance(JL.binding_ex(ctx3, binfo.id))
         is_from_user_ast(provs) || continue
         prov = last(provs)
+        if bk === :argument && kwarg_type_names !== nothing
```
aviatesk:
```suggestion
        if bk === :argument
```

PR #573, opened 2026-03-03
Correcting emacs configuration

yohad:
When using eglot, the usage of :autoport will cause jetls to not work properly when editing files over TRAMP. It can just be removed from the configuration and work just as well for both local and remote files.


yohad:
I hope what I wrote is okay

Issue #561, opened 2026-02-24
JETLS keeps running forever at 100% CPU (1 core) after closing Neovim


DoktorMike:
This comes from the related but curiously distinct from the VSCode issue linked here. 

See this comment for another neovim user @agdestein  with the same issue: https://github.com/aviatesk/JETLS.jl/issues/546#issuecomment-3944457481

> My setup is this:
> 
> OS: `Linux gauss 6.18.9-arch1-2 #1 SMP PREEMPT_DYNAMIC Mon, 09 Feb 2026 17:16:33 +0000 x86_64 GNU/Linux`
> JETLS: `JETLS version 2026-02-16`
> Nvim: `NVIM v0.11.6`
> Julia: `julia version 1.12.5`
> 
> My current config is this:
> 
> ```lua
>     config = function()
>       vim.lsp.config('jetls', {
>         cmd = {  '/home/mike/.julia/bin/jetls',  'serve', },
>         filetypes = { 'julia' },
>         root_markers = { 'Project.toml', 'JuliaProject.toml' },
>         settings = {},
>       })
>       vim.lsp.enable 'jetls'
> ```
> 
> where I use the newly recommended way of starting the server with `serve`.  So the command that actually get's run and which lingers after I close neovim is this:
> 
> ```bash
> julia --startup-file=no --startup-file=no --history-file=no --threads=auto -m JETLS serve
> ```
>  

 _Originally posted by @DoktorMike in [#546](https://github.com/aviatesk/JETLS.jl/issues/546#issuecomment-3943124712)_

DoktorMike:
Let me know if there's anything I can do to help debug this. I'm not very skilled in the language server protocol so the code is a bit opaque to me. I did try to have Claude hammer at it but it didn't solve the issue when I tested it locally.

benkj:
I'm having the same problem

ampalski:
Best I can tell I'm having the same issue. The command that lingers for me is:
`~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/bin/julia --startup-file=no --startup-file=no --history-file=no --threads=auto --threads=auto -m JETLS`

Config is
```
vim.lsp.config("jetls", {
	cmd = {
		"jetls",
		"--threads=auto",
		"--",
	},
	filetypes = { "julia" },
})
vim.lsp.enable("jetls")
```

abap34:
Would the problem still occur even if `exit_timeout` is configured?

```lua
[...]

PR #580, opened 2026-03-08
Fix hang on abrupt client death

abap34:
## Summary

Fix a bug where the JETLS process remains running indefinitely when a client such as Neovim terminates without sending the `exit` notification.

Two hang points in `Endpoint` are addressed.

## Background

In some setups, the client process appears to terminate before the language server returns the ShutdownResponse. This behavior has been observed with Neovim under their default configuration.

In such cases, JETLS is expected to shut down gracefully by monitoring the process Id.

However, two issues in `Endpoint` prevented this from happening and caused the server process to remain running indefinitely.

## Where the server hangs

```mermaid
sequenceDiagram
    participant N as Neovim
    participant R as read_task
    participant S as Server loop<br/>(calls iterate)
    participant W as write_task
    participant M as processId monitor

    N->>S: ShutdownRequest
    S->>W: ShutdownResponse (via out_msg_queue)
    N-xN: Process exits (no exit notification)
    Note over N: stdin/stdout/stderr pipes break

    R-->>R: readlsp returns nothing
    Note over R: Loop exits

    W-->>W: writelsp hits EPIPE
    W-->>W: err_handler writes to stderr
    Note over W: stderr also broken<br/>-> double EPIPE<br/>-> task dies

    S->>S: take!(in_msg_queue)
    Note over S: ❌ Hang 1<br/>Queue still open, blocks forever

    M->>M: Polls every 60s
    M->>S: SelfShutdownNotification

    Note over S: Loop exits -> finally

    S->>S: close(endpoint) -> flush()
    Note over S: ❌ Hang 2<br/>isready stays true (write_task dead)<br/>yield() loops forever
````

*(Both Mermaid charts were initially generated by Claude; I have manually reviewed them for correctness.)*

### Hang 1: `take!(in_msg_queue)` blocks

When stdin closes, `read_task` exits its loop, but `in_msg_queue` remains open. As a result, `take!(in_msg_queue)` in `iterate` keeps waiting for a message and blocks indefinitely.

The processId monitor (polling every 60 seconds) eventually resolves this by putting `SelfShutdownNotification` into the queue, but the server still fails to terminate due to Hang 2.

### Hang 2: `flush(endpoint)` loops indefinitely

`write_task` encounters EPIPE when writing to stdout. The code attempts to recover (`catch` -> `err_handler` -> `continue`), but the default `err_handler` writes to stderr using `@error` and `display_error`.

After the client exits, stderr is also a broken pipe. As a result, `err_handler` itself throws EPIPE, which is not caught and terminates `write_task`.

After the server loop exits, `finally` calls `close(endpoint)` -> `flush(endpoint)`. The `flush` implementation waits for the queue to drain:

```
while isready(out_msg_queue)
    yield()
end
```

However, `isready` only checks whether the queue contains data; it does not consider whether a consumer is still alive. Since `write_task` has already terminated, the queue never drains and the loop runs forever. This appears to be one of the main reasons the server hangs.

### Commit 1: Send a sentinel to `in_msg_queue`

When `read_task` exits, it inserts `nothing` into `in_msg_queue`.

`iterate` detects this sentinel and returns `nothing`, which ends the server loop and proceeds to the `finally` block.

### Commit 2: Detect `write_task` termination in `flush`

The `flush` loop now checks

```
istaskdone(endpoint.write_task)
```

and exits when the consumer task has already terminated. This allows `close(endpoint)` in the `finally` block to complete and the server process to exit normally.

The underlying cause of Hang 2 is the unintended termination of `write_task` due to the double EPIPE in `err_handler`. That issue could be addressed separately by fixing `err_handler`. This change instead ensures robustly that `flush` does not hang regardless of why `write_task` terminates.

## After the fix

Before the fix, the process would remain active every time, but now it closes immediately once the editor is shut down!

```mermaid
sequenceDiagram
    participant N as Neovim
    participant R as read_task
    participant S as Server loop<br/>(calls iterate)
    participant W as write_task

    N->>S: ShutdownRequest
    S->>W: ShutdownResponse (via out_msg_queue)
    N-xN: Process exits (no exit notification)
    Note over N: stdin/stdout/stderr pipes break

    R-->>R: readlsp returns nothing
    Note over R: Loop exits

    R->>S: put!(in_msg_queue, nothing)
    Note over R: ✅ Fix 1: sentinel

    W-->>W: double EPIPE -> task dies

    S->>S: take! returns nothing
    Note over S: Loop exits -> finally

    S->>W: istaskdone? -> true
    Note over S: ✅ Fix 2: flush breaks

    Note over S: Process exits normally
```

## Related issue

This may resolve aviatesk/JETLS.jl#561, but there could be other contributing factors, so it may be better to wait before closing it. For example, I noticed that the process can remain when Neovim is closed immediately after startup (this does not seem to occur with other editors).

## Notes

Testing every possible scenario is quite challenging, so this has not been verified comprehensively. However, after implementing the fix and using it for a while with Neovim, Helix, and VSCode, I confirmed that processes start, communicate, and terminate correctly.

Feedback from Neovim & other editors would be appreciated.

### DIFF
--- a/LSP/src/communication.jl
+++ b/LSP/src/communication.jl
@@ -36,10 +36,10 @@ close(endpoint)
 ```
 """
 mutable struct Endpoint
-    in_msg_queue::Channel{Any}
-    out_msg_queue::Channel{Any}
-    read_task::Task
-    write_task::Task
+    const in_msg_queue::Channel{Any}
+    const out_msg_queue::Channel{Any}
+    const read_task::Task
+    const write_task::Task
     @atomic isopen::Bool
 
     function Endpoint(err_handler, in::IO, out::IO)


abap34:
@aviatesk Does this PR work correctly in Zed? I tried it in Helix for a while, and it does not seem to cause any major issues. It also appears that the original problem has been fixed, so it might be fine to merge it.

aviatesk:
Thanks for the thorough analysis and the fix — I strongly agree with the overall approach. Let's move forward with this.

I've confirmed that the normal server lifecycle (Zed, VSCode) is not affected by these changes.

Since this code is fairly subtle, I'll push a few additional comments and safeguards for future reference. Once I push those, could you take a look and confirm? Then we can merge.

LSP/src/communication.jl
```diff
@@ -48,16 +48,23 @@ mutable struct Endpoint
...
+                end break # terminate this task loop when the stream is closed
+                (!isassigned(endpoint_ref) || isopen(endpoint_ref[])) || break
+                put!(in_msg_queue, msg)
+                GC.safepoint()
+            end
+            # Send a sentinel to unblock `take!` in `iterate` — without this,
+            # the server loop hangs forever when the input stream closes.
+            # Guard with `isopen` since `close(endpoint)` may have already
+            # closed the channel during normal shutdown.
+            isopen(in_msg_queue) && put!(in_msg_queue, nothing)
```
aviatesk, line 67: When `read_task` exits during normal shutdown, `close(endpoint)` may have already closed `in_msg_queue`. Without this guard, `put!` might throw `InvalidStateException` on the closed channel. The exception is silently swallowed (since `read_task` is never `wait`ed on), but the guard makes the intent explicit and avoids the unnecessary exception.

LSP/src/communication.jl
aviatesk: I started to think this call to `flush` and the `flush` overload actually is redundant — `close` puts a `nothing` termination token into `out_msg_queue` and then calls `wait(write_task)`, which already guarantees that all prior messages are processed in FIFO order. We could remove `flush` and its call in `close` entirely. We may try to remove it entirely in this PR, or at least it is worth discussing separately.

aviatesk:
Also, please update CHANGLEOG.md. This is an important enhancement.

LSP/src/communication.jl
aviatesk: I added a commit to remove this entirely: [234192c](https://github.com/aviatesk/JETLS.jl/pull/580/commits/234192c50f5a79265cb3cc2eab00ae8fe7ae40d4)

I confirmed this change doesn't break any existing server lifecycle management in Zed/VSCode, but I'd appreciate if you could check this on Neovim/Helix too.

aviatesk (approved):
Thanks so much for working on this, @abap34 !

PR #615, opened 2026-04-01
feat : field completions

Sypher845:
Kooha-2026-04-02-04-49-59

This PR adds field completion support to JETLS. When a user types `x.` and triggers completions (e.g., via `Ctrl+Space`), the server now suggests available fields for the inferred type of `x`.

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -4,6 +4,7 @@
 const NUMERIC_CHARACTERS = tuple(string.('0':'9')...)
 const METHOD_COMPLETION_TRIGGER_CHARACTERS = ("(", ",", " ")
 const COMPLETION_TRIGGER_CHARACTERS = [
+    ".",  # dot completion
     "@",  # macro completion
     "\\", # LaTeX completion
     ":",  # emoji completion
@@ -142,7 +143,10 @@ end
 
 should_invoke_auto_completion(::Nothing, ::Bool=false) = true
 function should_invoke_auto_completion(context::CompletionContext, allow_macro::Bool=false)
-    if !allow_macro || context.triggerCharacter != "@"
+    allow_character_trigger =
+        context.triggerCharacter == "." ||
+        (allow_macro && context.triggerCharacter == "@")
+    if !allow_character_trigger
         # Don't trigger completion just by typing a numeric character, etc.
         if context.triggerKind != CompletionTriggerKind.Invoked
             return false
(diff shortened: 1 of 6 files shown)


aviatesk:
This PR implements `.`-completion as "field completion", but in Julia, `x.y` is actually not a field access but it's a property access. That means `x.`-completion needs to be implemented in a way that emulates the behavior of `getproperty(::typeof(x), ::Symbol)`. The current implementation calls `fieldnames` on the inferred type, which is the most reliable but incomplete way to statically emulate `getproperty`. Do you have any ideas on this?

Also, there should be additional points to discuss:
- `ASTTypeAnnotator` probably expects a complete AST in most cases. However, code like `xxx.|` might produce an incomplete AST. I'd like you to investigate what happens in which cases, and check whether the current approach still works fine.
- Please improve the completion trigger logic so that `.|` automatically triggers completion.

Lastly, please rebase this PR and set aviatesk/JETLS.jl#620 as the merge target.

src/completions.jl
```diff
@@ -174,6 +174,44 @@ function local_completions!(
...
+    T isa DataType || return #=isIncomplete=#false
+    isabstracttype(T) && return #=isIncomplete=#false
+    fnames = @invokelatest fieldnames(T)
+    for (i, fname) in enumerate(fnames)
+        s = String(fname::Symbol)
+        ftyp = @invokelatest fieldtype(T, i)
+        items[s] = CompletionItem(;
+            label = s,
+            labelDetails = CompletionItemLabelDetails(;
+                detail = "::" * string(ftyp),
```
aviatesk: `string(ftyp)` needs to be "postprocess"ed, since the actual type instance may be defined within the JET's virtual module.

src/hover.jl
```diff
@@ -55,15 +55,35 @@ function local_binding_hover_info(fi::FileInfo, uri::URI, definitions::JS.Syntax
     return String(take!(io))
 end
 
+function type_hover_for_binding(
```
aviatesk: Don't include the type on hover changes in this PR.

Sypher845:
@aviatesk i have addressed the main review points,
- `. `now triggers completion automatically.
- dot completion now handles properties, not only fields.
- typed local dot completion still works, including incomplete cases like `p. `.
- displayed type details are postprocessed before showing them.

Current limitation:
for inferred non-constant values, custom property completion currently works for simple static `propertynames(::T) `cases, but not for all dynamic `getproperty `cases.

Please take a look when you have time

aviatesk:
@Sypher845 consolidated feedback before we iterate further — the concrete items from the first review are addressed (`.` trigger, `propertynames` framing, post-processing, tests), thanks for that.

The remaining concern is approach-level. The current PR layers three parallel paths — `add_dot_property_completions!`, `add_static_dot_property_completions!`, `add_dot_field_completions!` — that are each a partial, ad-hoc emulation of `propertynames` over a different slice of inputs. What I'd have liked you to notice on your own is that building several incomplete emulations in parallel isn't a good way to approach a problem like this; it's a sign that the single underlying question, *"what does `propertynames(x)` return in this context?"*, should be answered once by the right tool rather than approximated three times.

The unified formulation is: obtain the inferred type (or `Core.Const` value) of the dot prefix — `ASTTypeAnnotator` gives you this for local bindings — then abstract-call `propertynames(::T)` through an `AbstractInterpreter` (`LSAnalyzer` is sufficient for this part). If the result is a `Core.Const(::Tuple{Vararg{Symbol}})`, use those names directly. The type detail for each name can come from similarly abstract-calling `getproperty(::T, Core.Const(:name))`. Three paths collapse into one, and the "is `propertynames` overridden?" question disappears — inference resolves both cases uniformly.

One way to have caught this earlier is to try to break an approach like the current one on a non-trivial example, e.g. `Regex`:

```julia
julia> code_typed((Regex,); optimize=false) do r
           propertynames(r)
       end
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Main.propertynames::Core.Const(propertynames)
│   %2 =   dynamic (%1)(r)::Core.Const((:pattern, :compile_options, :match_options, :regex))
└──      return %2
) => NTuple{4, Symbol}
```

The current `collect_static_property_names` cannot reach this case — the `length(stmts) == 2` assumption rejects branching bodies — whereas the abstract-call approach handles it uniformly, and as a side benefit avoids deserializing IR on each completion.

If you'd like to take this forward, the path is to rework the implementation around this abstract-call design.

Sypher845:
Thanks for the detailed breakdown. I understand the issue now.  I'll look into this and try to rework the implementation around that design. Will update once I have something concrete.

aviatesk:
xref: https://github.com/aviatesk/JETLS.jl/pull/698

PR #633, opened 2026-04-16
Use MarkupContent in diagnostic message if supported by client

jwortmann:
I noticed that most diagnostic messages from JETLS use inline code in standard Markdown format (delimited by backticks).

The upcoming 3.18 version of the LSP specs allows to send [diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnostic) messages with MarkupContent if the client has support for that:
```ts
	/**
	 * The diagnostic's message.
	 *
	 * @since 3.18.0 - support for MarkupContent. This is guarded by the client
	 * capability `textDocument.diagnostic.markupMessageSupport`.
	 */
	message: string | MarkupContent;
```

This PR proposes to use MarkupContent in diagnostic messages if applicable for the message and if supported by the client. This allows the client to render the message in a nicer way.

The following screenshots are from Sublime Text and with a currently unreleased version of the LSP client, which has recently added support for that (it will be available with the next release of the LSP package for Sublime Text).

Before:

before

---

After:

after

---

I want to mention that technically the client capability for `markupMessageSupport` is only part of the [DiagnosticClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#diagnosticClientCapabilities) (`textDocument.diagnostic.markupMessageSupport`), and not part of the [PublishDiagnosticsClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#publishDiagnosticsClientCapabilities) (`textDocument.publishDiagnostics`). However, I believe this is just a small inconsistency in the LSP specs, and it wouldn't really make sense for the client to limit this to only pull diagnostics, so I also applied the MarkupContent to diagnostics from `publishDiagnostics` notifications. The docstring for the `Diagnostic` type that I quoted above also refers only to that single capability path, so I assume that it should be fine.

### DIFF
--- a/src/diagnostic.jl
+++ b/src/diagnostic.jl
@@ -256,6 +256,17 @@ function diagnostic_code_description(code::AbstractString)
         href = URI("https://aviatesk.github.io/JETLS.jl/release/diagnostic/#diagnostic/reference/$code"))
 end
 
+function apply_markdown_message!(diagnostics::Vector{Diagnostic})
+    i = 1
+    while i <= length(diagnostics)
+        diagnostic = diagnostics[i]
+        if diagnostic.message isa String && occursin('`', diagnostic.message)
+            diagnostics[i] = Diagnostic(diagnostic; message = MarkupContent(; kind = MarkupKind.Markdown, value = diagnostic.message))
+        end
+        i += 1
+    end
+end
+
 # utilities
 # =========
 
@@ -1626,6 +1637,9 @@ function notify_diagnostics!(server::Server, uri2diagnostics::URI2Diagnostics; e
             continue
         end
         apply_diagnostic_config!(diagnostics, state.config_manager, uri, root_path)
+        if supports(server, :textDocument, :diagnostic, :markupMessageSupport)
+            apply_markdown_message!(diagnostics)
+        end
         send(server, PublishDiagnosticsNotification(;
             params = PublishDiagnosticsParams(;
                 uri,
@@ -1746,6 +1760,9 @@ function handle_DocumentDiagnosticRequest(
     if notebook_uri !== nothing
         diagnostics = localize_notebook_diagnostics(server.state, notebook_uri, uri, diagnostics)
     end
+    if supports(server, :textDocument, :diagnostic, :markupMessageSupport)
+        apply_markdown_message!(diagnostics)
+    end
     return send(server,
         DocumentDiagnosticResponse(;
             id = msg.id,
@@ -1826,6 +1843,10 @@ function send_workspace_diagnostics(
         notebook_uri = get_notebook_uri_for_cell(state, uri)
         if notebook_uri !== nothing
             diagnostics = localize_notebook_diagnostics(state, notebook_uri, uri, diagnostics)
+        end
+
+        if supports(server, :textDocument, :diagnostic, :markupMessageSupport)
+            apply_markdown_message!(diagnostics)
         end
 
         item = WorkspaceFullDocumentDiagnosticReport(;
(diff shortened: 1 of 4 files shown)


src/diagnostic.jl
```diff
@@ -256,6 +256,17 @@ function diagnostic_code_description(code::AbstractString)
...
+function apply_markdown_message!(diagnostics::Vector{Diagnostic})
+    i = 1
+    while i <= length(diagnostics)
+        diagnostic = diagnostics[i]
+        if diagnostic.message isa String && occursin('`', diagnostic.message)
+            diagnostics[i] = Diagnostic(diagnostic; message = MarkupContent(; kind = MarkupKind.Markdown, value = diagnostic.message))
+        end
+        i += 1
+    end
+end
```
aviatesk, line 265: Using a simple `for i = 1:length(diagnostics)` loop would be cleaner here, I think.

src/diagnostic.jl
```diff
@@ -256,6 +256,17 @@ function diagnostic_code_description(code::AbstractString)
         href = URI("https://aviatesk.github.io/JETLS.jl/release/diagnostic/#diagnostic/reference/$code"))
 end
 
+function apply_markdown_message!(diagnostics::Vector{Diagnostic})
+    i = 1
+    while i <= length(diagnostics)
+        diagnostic = diagnostics[i]
+        if diagnostic.message isa String && occursin('`', diagnostic.message)
```
aviatesk:
```suggestion
        if diagnostic.message isa String
```
Markdown syntax isn't limited to backticks, so we could skip this condition and just replace all the diagnostics with the ones that use the `MarkupContent` version. Since `Diagnostic` is an immutable struct, the performance cost should be negligible.

aviatesk (commented):
Nice improvement. I hadn't looked into LSP 3.18 details yet, but I think clients will gradually add support for it, so I really appreciate you making this change.

I left a few comments.

To merge this, please also fix the selfcheck errors (you'll need to modify `cli-check.jl`): https://github.com/aviatesk/JETLS.jl/actions/runs/24489449489/job/71571460604?pr=633
To run the selfcheck locally, you can just run `./scripts/selfcheck` from the root of this repo.
Don't worry about the vendor-deps CI failure.

src/diagnostic.jl
jwortmann: Yes, but this also carries a cost at the client side, which needs to pass diagnostic messages into a Markdown parser in that case. For the example from the screenshot showing a popup on mouse hover this should be negligible, but in Sublime Text we also have an alternative mode to permanently display diagnostic messages on the right side of the line directly in the editor:

diagnostics

In that case it could be useful to skip the Markdown parsing if not necessary.

Do you know that you actually use other Markdown features than the inline code in some of the diagnostic messages?

src/diagnostic.jl
jwortmann: https://github.com/aviatesk/JETLS.jl/pull/633/changes/a9c6386ebf0691d52d103ace77755b3f6d3f43cd

src/diagnostic.jl
aviatesk: There aren't many, but some of the top-level error messages returned by full-analysis do contain things like bullet points. Those would be better rendered as markdown.
More of these might be added in the future, and it's a bit inconvenient to have to keep this check in mind every time we add new diagnostics.

The performance concern for inline diagnostics is valid, but since the capability is being declared, it feels unnatural to keep code on the server side that's driven by client-side performance concerns.
I think a better solution would be for Sublime to just skip markdown rendering entirely for inline diagnostics?

jwortmann: Okay, should be fine I guess. In case it turns out to have performance implications for inline diagnostics, I can take action on the client side.

aviatesk (approved):
Thanks so much for this enhancement.

PR #645, opened 2026-04-21
tests: tighten ASTTypeAnnotator boundaries

Sypher845:
## Summary

This PR improves `ASTTypeAnnotator` and adds focused coverage for it.

## What changed

- fix top-level declaration handling so it does not poison inference
- add real-world tests
- add incomplete-syntax boundary tests
- include the new test file in `test/runtests.jl`

## Notes

This PR is scoped to `ASTTypeAnnotator` only.

## Testing

- ran the focused `ASTTypeAnnotator` test file

### DIFF
--- a/src/analysis/ASTTypeAnnotator.jl
+++ b/src/analysis/ASTTypeAnnotator.jl
@@ -1,0 +1,313 @@
+using Core.IR
+
+struct ASTTypeAnnotatorToken end
+
+struct ASTTypeAnnotator <: CC.AbstractInterpreter
+    toptree::JL.SyntaxTree
+    topmi::Core.MethodInstance
+    limit_aggressive_inference::Bool
+    world::UInt
+    inf_params::CC.InferenceParams
+    opt_params::CC.OptimizationParams
+    inf_cache::Vector{CC.InferenceResult}
+    function ASTTypeAnnotator(
+            toptree::JL.SyntaxTree,
+            topmi::Core.MethodInstance,
+            limit_aggressive_inference::Bool = false;
+            world::UInt = Base.get_world_counter(),
+            inf_params::CC.InferenceParams = CC.InferenceParams(;
+                aggressive_constant_propagation = true
+            ),
+            opt_params::CC.OptimizationParams = CC.OptimizationParams(),
+            inf_cache::Vector{CC.InferenceResult} = CC.InferenceResult[]
+        )
+        return new(toptree, topmi, limit_aggressive_inference, world, inf_params, opt_params, inf_cache)
+    end
+end
+CC.InferenceParams(interp::ASTTypeAnnotator) = interp.inf_params
+CC.OptimizationParams(interp::ASTTypeAnnotator) = interp.opt_params
+CC.get_inference_world(interp::ASTTypeAnnotator) = interp.world
+CC.get_inference_cache(interp::ASTTypeAnnotator) = interp.inf_cache
+CC.cache_owner(::ASTTypeAnnotator) = ASTTypeAnnotatorToken()
+
+# ASTTypeAnnotator is only used for type analysis, so it should disable optimization entirely
+CC.may_optimize(::ASTTypeAnnotator) = false
+
+# ASTTypeAnnotator doesn't need any sources to be cached, so discard them aggressively
+CC.transform_result_for_cache(::ASTTypeAnnotator, ::CC.InferenceResult, ::Core.SimpleVector) = nothing
+
+# ASTTypeAnnotator analyzes a top-level frame, so better to not bail out from it
+CC.bail_out_toplevel_call(::ASTTypeAnnotator, ::CC.InferenceLoopState, ::CC.InferenceState) = false
+
+# `ASTTypeAnnotator` aggressively resolves global bindings to enable reasonable completions
+# for lines like `Mod.a.|` (where `|` is the cursor position).
+# Aggressive binding resolution poses challenges for the inference cache validation
+# (until https://github.com/JuliaLang/julia/issues/40399 is implemented).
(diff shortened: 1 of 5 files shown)


test/analysis/test_ASTTypeAnnotator.jl
```diff
@@ -0,0 +1,122 @@
...
+
+    let st0 = jlparse("x."; rule=:statement)
+        @test JS.kind(st0) === JS.K"."
+        @test JS.numchildren(st0) ≥ 2
+        @test JS.kind(st0[1]) === JS.K"Identifier"
+        @test JS.kind(st0[2]) === JS.K"inert"
+    end
+    @test_throws JL.LoweringError lower_statement("x.")
+    @test_throws JL.LoweringError lower_statement("Base.")
+    @test_throws JL.LoweringError lower_statement("foo(; a=")
```
aviatesk: I don't understand why these cases are tested here.. Nothing being tested here seems to be involved with ASTTypeAnnotator.

src/analysis/ASTTypeAnnotator.jl
```diff
@@ -206,8 +216,13 @@ end
...
     )
     if ret isa CC.AbstractEvalBasicStatementResult
+        rt = ret.rt
+        # Keep Core top-level declarations from poisoning inference with `Union{}`
+        if rt === Union{} && is_core_toplevel_declaration_call(stmt)
+            rt = Nothing
+        end
         ret = CC.AbstractEvalBasicStatementResult(
-            ret.rt, ret.exct, ret.effects, ret.changes, ret.refinements,
+            rt, ret.exct, ret.effects, ret.changes, ret.refinements,
```
aviatesk, line 295: Why is this change necessary?

test/analysis/test_ASTTypeAnnotator.jl
```diff
@@ -0,0 +1,122 @@
...
+function query_source_type(st0::JS.SyntaxTree, tree::JS.SyntaxTree, text::AbstractString)
+    node = only_node_with_text(st0, text)
+    return JETLS.get_type_for_range(tree, JS.byte_range(node))
+end
+
+@testset "basic inference" begin
+    let (; st0, tree) = infer_statement("sin(1.0)")
+        @test tree !== nothing
+        typ = JETLS.get_type_for_range(tree, JS.byte_range(st0))
+        @test typ === Core.Const(0.8414709848078965)
```
aviatesk:
```suggestion
        @test typ === sin(1.0)
```

src/analysis/ASTTypeAnnotator.jl
Sypher845: I checked this on lowered code: the base evaluator returns `Union{}` for `Core.declare_global` and `Core.declare_const`. Since ASTTypeAnnotator annotates statements from` frame.src.ssavaluetypes`, those declaration statements would otherwise be annotated as Union{}. I normalize only those bookkeeping statements to Nothing.

test/analysis/test_ASTTypeAnnotator.jl
Sypher845: will just remove them

test/analysis/test_ASTTypeAnnotator.jl
Sypher845: `get_type_for_range` returns `Core.Const(...) ` here, not the raw value, so I changed it to `Core.Const(sin(1.0))`

Sypher845:
@aviatesk PTAL, i have addressed review comments.

src/analysis/ASTTypeAnnotator.jl
aviatesk: I know those are inferred as `Union{}`, but do you understand "why" that happens? Given that JETLS currently only supports v1.12, it might be unavoidable to keep this code in JETLS, but even so, I think we should add a comment explaining why this is needed.

Sypher845: Yes. I added a comment explaining why this is needed.

aviatesk: Please enable this code only on versions that need this.

test/analysis/test_ASTTypeAnnotator.jl
```diff
@@ -0,0 +1,112 @@
...
+        @test CC.widenconst(typ) === Float64
+    end
+
+    let (; st0, tree) = infer_statement("(Ref(Some(42))[]).value")
+        @test tree !== nothing
+        typ = JETLS.get_type_for_range(tree, JS.byte_range(st0))
+        @test typ !== nothing
+        @test CC.widenconst(typ) === Int64
+    end
+end
```
aviatesk, line 116: What should really be tested is whether each identifier, expression, and so on in examples like this gets an appropriate type assigned. For example, in the first example, does `onfig.scale` get the right type?
I think it's fine to also test the return type of this whole code as these tests do currently, but that isn't the main point of `ASTTypeAnnotator`.

And we should also test that it works for things like function definitions.

Sypher845: Done , i added tests for testing if each identifier, expression, etc gets there type.
For the function definitions, i had to do some changes to the Annotator, so now both the fine grained type assigns and function definitions  also works.

aviatesk:
Thanks for the contribution. Since this PR was opened, I've significantly redesigned the `ASTTypeAnnotator`: the method-lookup-based approach can't reliably resolve gensymed names from keyword-method and closure bodies. The new design infers each method body as an anonymous toplevel chunk, side-stepping the synthetic-name issue entirely.

Closing in favor of #620, where the rework — including a fresh test suite — is being completed.

Issue #723, opened 2026-05-28
diagnostic Confusing error message from `lowering/unused-assignment`.


EdsterG:
### Pre-submission checklist

- [x] I have searched [existing issues](https://github.com/aviatesk/JETLS.jl/issues) and confirmed this is not a duplicate.

### Client

ST4

### Julia `versioninfo()`

```text
Julia Version 1.12.2
Commit ca9b6662be4 (2025-11-20 16:25 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
```

### JETLS version

2026-05-27

### Reproduction with the VSCode reference client

- [ ] I have confirmed this issue also reproduces with the VSCode reference client.
- [ ] This issue does NOT reproduce with the VSCode reference client, but I believe JETLS-side changes are needed.

### Description

When a variable is implicitly returned it is marked as "unused". Since Julia, by default, returns the last expression I would expect JETLS to respect Julias default behavior and assume the variable is used because is it returned.

### Steps to reproduce

Here's the minimum example to reproduce:
```
function test(x::String)
    str = "($x)"
end
```
Here's a more practical example, where a local variable is used for incremental string construction:
```
function test(x)
    str = "("
    for i in x
        str *= x
        str *= ","
    end
    str *= ")"
end
```

aviatesk:
I think the current implementation feels natural.

In your example, the last assigned `str` is not actually used. Julia returns the value of the last expression in a function, and in your example, the last expression is an assignment, and the value of an assignment expression is the value of its RHS. And the LHS variable that receives the assigned value is never used.

In this case, I think it is semantically natural for the program to have `(return) str` at the end of the function.

aviatesk:
I think the current implementation feels natural.

In your example, the last assigned `str` is not actually used. Julia returns the value of the last expression in a function, and in your example, the last expression is an assignment, and the value of an assignment expression is the value of its RHS. And the LHS variable that receives the assigned value is never used.

In this case, I think it is semantically natural for the program to have `(return) str` at the end of the function.

EdsterG:
I understand your reasoning. This syntax is a feature of the language, so it's ultimately a question of style. Since you're the author of this great tool you can enforce whatever style you'd like, but to me it seems like this warning is in opposition with a feature of the language.

aviatesk:
What do you think on https://github.com/aviatesk/JETLS.jl/pull/727?

[...]

PR #727, opened 2026-06-02
diagnostics: Clarify tail assignment diagnostics

aviatesk:
Address aviatesk/JETLS.jl#723.

In tail-position assignments, Julia returns the assignment expression, while the assigned binding itself may still never be read.
For example, in:
```julia
function f()
    x = 1
end
```
the function returns the value of `x = 1`, not a later read of `x`. 
The old `Unused local binding x` wording did not explain that implicit-return distinction. 

JETLS now reports this message, wrapped here:
```
Local binding `x` is not read; consider `return x` to return it explicitly
```

For tail dead stores, JETLS similarly reports a message of the form, wrapped here:
```
Value assigned to `x` is returned directly; consider `return x` to return the binding explicitly
```

Classify returned assignment expressions and report this more specific message. Simple block-tail assignments also carry insertion metadata so the quick fix can add that explicit return.

### DIFF
--- a/LSP/src/basic-json-structures.jl
+++ b/LSP/src/basic-json-structures.jl
@@ -623,6 +623,8 @@ struct UnusedVariableData
     is_tuple_unpacking::Bool
     assignment_range::Union{Nothing,Range}
     lhs_eq_range::Union{Nothing,Range}
+    return_insert_position::Union{Nothing,Position}
+    return_insert_text::Union{Nothing,String}
 end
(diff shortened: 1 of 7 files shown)


EdsterG:
Thank you for this change. I think the word "implicitly" would better describe the behavior instead of "directly":
```
Value assigned to `x` is returned implicitly; consider `return x` to make the return explicit.
```

PR #782, opened 2026-06-27
Support Julia 1.14-DEV

koalazub:
JETLS does not load or analyse on a Julia 1.14 nightly. The dev Compiler has reworked the internals a language server necessarily reaches into, so the points where JETLS reads inference state, inspects call information, and drives the type annotator no longer match what the running compiler offers.

The change teaches those call sites to ask the compiler what shape it is in rather than assume the one they were written against. Every adaptation is feature detected and resolved at parse time, so a 1.12 or 1.13 build lowers exactly as it does today and only a nightly compiles the new arms. The goal is a single tree that serves the released compilers and the moving one together, instead of a long-lived branch that drifts further apart with every nightly.

A second, unrelated problem is fixed alongside it. Notebook cells that hold prose rather than code, written as `@markdown`, `@raw`, and `@typst`, were still handed to the Julia parser and came back as syntax and macro errors. Those cells are now masked before parsing while their raw text is kept for formatting, so diagnostics describe the code someone actually wrote.

This needs the companion Compiler support in aviatesk/JET.jl#836 to be useful.

### DIFF
--- a/src/JETLS.jl
+++ b/src/JETLS.jl
@@ -396,7 +396,7 @@ function handler_concurrent_message(server::Server, @nospecialize msg)
         end
     elseif isdefined(msg, :id) && (id = msg.id; id isa String || id isa Int)
         let cancel_flag = get!(()->CancelFlag(false), server.state.currently_handled, id)
-            Threads.@spawn :default @tryinvokelatest handle_request_message(server, msg, cancel_flag)
+            Threads.@spawn :default @tryinvokelatest handle_request_message_or_respond_error(server, msg, cancel_flag)
         end
     else
         Threads.@spawn :default @tryinvokelatest handle_notification_message(server, msg)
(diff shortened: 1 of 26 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@ out
 Project.toml.bak
 Manifest.toml
 Manifest-*.toml
+LocalPreferences.toml
```
aviatesk, line 6: This is such a simple change that can be merged separately.

koalazub: Agreed — split out into #783 so it can be merged on its own. I've also removed it (and the rest of the incidental local-dev scaffolding: the `[sources]` path, `LocalPreferences.toml`, and the `[compat]` churn) from this PR, so #782 is now just the 1.14 Compiler/JET-API adaptations.

koalazub:
The nightly failure is upstream of this diff — the registered Compiler package is stale against the current nightly (`UndefVarError: _eval_import` at precompile), so JET/JETLS never compile.

PR #783, opened 2026-06-27
Ignore generated LocalPreferences.toml

koalazub:
`LocalPreferences.toml` is generated by Preferences.jl for local, machine-specific package configuration and should not be tracked. Split out of #782, where it rode along incidentally with the 1.14 work, as suggested in review so it can be merged on its own.


aviatesk:
Thanks!

PR #805, opened 2026-07-27
full-analysis: Remove configurable analysis worker count

aviatesk:
The experimental `n_analysis_workers` option allowed multiple full-analysis requests to run concurrently. Outside signature analysis, full analysis primarily performs environment-dependent code loading, where JETLS must serialize that work to keep the process-wide active project stable. JET virtual processing also mutates process-global state and is not safe to run concurrently.

On the other hand, the CPU-intensive signature-analysis phase is already parallelized across the available Julia threads. Multiple full-analysis workers (which can be configured by `n_analysis_workers`) therefore provide little additional useful parallelism while adding scheduling and lifecycle complexity.

If request-level parallelism is needed in the future, it should be implemented with isolated analysis processes rather than additional in-process workers. Process isolation would also let JETLS restart only the user-code process when Revise cannot restore a consistent state.

So this commit processes full-analysis requests through one worker, store its task in a single assigned `Ref{Task}`, and remove `n_analysis_workers` from initialization parsing, generated schemas, documentation, and the VSCode client.

This is a breaking configuration change. Existing users must remove `n_analysis_workers` from client or `.JETLSConfig.toml` settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Breaking
 
+- Removed the experimental `n_analysis_workers` initialization option. Full-analysis requests are now processed serially, while signature analysis within each request remains parallel based on the available Julia threads. Existing client or `.JETLSConfig.toml` configurations must remove this option.
+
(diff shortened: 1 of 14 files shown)


aviatesk:
"Self diagnostics" is hanging due to a known issue.
This will be fixed in a subsequent PR (in fact, this PR was submitted as preparation for that fix).
xref: #806

PR #831, opened 2026-08-04
full-analysis: Add configurable concretization patterns

### LINKED ISSUE #464: Global variable is not concretized
To reproduce:
```
git clone https://github.com/ufechner7/FLORIDyn.jl
cd FLORIDyn.jl
```
and open the file `examples/mpc_162.jl`.

I get lots of errors like this one:
```
`MAX_STEPS` is not concretized but JET needs to use its actual value in order to define types or methods.
- If this binding can be declared as a constant, try to declare it as a constant (i.e. `const MAX_STEPS = ...`).
- You may need to specify `:(MAX_STEPS = x_)` pattern to the `concretization_patterns`
  configuration to allow JET to actually evaluate this binding, e.g.,
  `report_file("path/to/file.jl"; concretization_patterns = [:(MAX_STEPS = x_)])`.
- If the above approaches do not work, try `concretization_patterns = [:(x_)]` to
  concretize all top-level code in the module (recommended as a last resort since it
  would incur any side effects in your code and may cause the analysis to take longer time).
```
I can get rid of this if I define MAX_STEPS as constant, but that is not always possible.

There is a second suggestion in the warning, but I cannot see how to follow it when using JETLS. My impression is, that the text of this warning refers to people who use JET.jl directly and not JETLS.jl. 

Is there a way to specify a concretization pattern for JETLS.jl, for example in `.JETLSConfig.toml` ?

aviatesk:
## Summary

JETLS previously reported missing top-level concrete values through the generic `toplevel/error` diagnostic, with remediation guidance that referred to JET's direct API and could not be configured through JETLS.

This PR adds a dedicated `toplevel/missing-concretization` diagnostic and introduces `full_analysis.concretization_patterns` for `.JETLSConfig.toml` and LSP settings.

```toml
[[full_analysis.concretization_patterns]]
pattern = "USE_PULSE = x_"
path = "src/config.jl"
```

Each entry accepts a MacroTools expression pattern and an optional workspace-relative path glob. Configured patterns supplement JET's built-in patterns and apply only to files inside the workspace.

The new `toplevel/missing-concretization` diagnostic provides a preferred quick fix that creates or updates `.JETLSConfig.toml`. The generated pattern preserves the relevant assignment syntax and is scoped to the file containing the assignment, which may differ from the file containing the diagnostic.

Changes to analysis-related configuration automatically trigger reanalysis, including when the initial analysis is still queued or running.

Matching top-level expressions are evaluated in the JETLS analysis process and may have side effects, so patterns should be kept as specific as possible.

## Additional changes

- Added configuration schemas and client metadata for the new setting.
- Added documentation for configuring patterns and using the quick fix.
- Added focused coverage for configuration parsing, diagnostics, code actions, path matching, TOML editing utilities (offering a support for existing `.JETLSConfig.toml` files that represent `concretization_patterns` as inline arrays.), and configuration-triggered reanalysis.

----

Closes https://github.com/aviatesk/JETLS.jl/issues/464

### DIFF
--- a/.JETLSConfig.toml
+++ b/.JETLSConfig.toml
@@ -66,6 +66,13 @@ match_type = "regex"
 severity = "off"
 path = "src/testrunner/testrunner-types.jl"
 
+[[full_analysis.concretization_patterns]]
+pattern = "const devbranch = x_"
+path = "docs/make.jl"
+[[full_analysis.concretization_patterns]]
+pattern = "const release_date = x_"
+path = "docs/make.jl"
+
 [initialization_options]
 reuse_native_inference = true
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Added a dedicated [`toplevel/missing-concretization`](https://aviatesk.github.io/JETLS.jl/release/diagnostic/#diagnostic/reference/toplevel/missing-concretization) diagnostic for top-level bindings whose concrete values JET needs during full analysis.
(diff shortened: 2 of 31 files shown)


ufechner7:
Thank you so much!

